PR #55, opened 2020-08-02
Issue 45: Disallow spaces in user names

jim-adventureprogames:
Front end changes, updates username automatically by turning ' ' into '-', and also removes ' ' on submit if somehow they snuck in.

### DIFF
--- a/server/src/user.ts
+++ b/server/src/user.ts
@@ -49,13 +49,31 @@ export async function getUserIdForUsername(username: string) {
 }
 
 export async function updateUserProfile(userId: string, data: Partial<User>) {
-  const profile = (await DB.getPublicUser(userId)) || {};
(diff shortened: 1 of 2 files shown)


lazerwalker:
Awesome, looks good!

There are some slight server-side changes necessary — basically, the /updateProfile endpoint should run the same replacement code just to make sure nobody's hand-crafted their own HTTP request and sidestepped our client-side validation.

I'm happy to do that work (basically just copy/pasting your function over and putting it in the right place) if you don't want to bother going through the work of setting up a working server environment, your call.

jim-adventureprogames:
I am not ready to set up a backend environment locally just yet.

lazerwalker:
Added this check to the server, rebased against the main branch, and merging this in. Thanks!

Issue #25, opened 2020-07-15
Find a good monospace typeface we can use, embed it.


lazerwalker:
We should pick our own distinctive beautiful monospace typeface we can embed within the app so everybody gets the same experience.

I think Fira Code is a good choice. Microsoft's Cascadia Code is too. Both are OSS.

We want something with bold and ideally italics.

kawa-kitsuragi:
I'm fond of Roboto Mono: https://fonts.google.com/specimen/Roboto+Mono It does have both italics and bold. Selfishly, I like its @ a little more than Fira's, too! 

Surfing Google Fonts also had me find this, which intrigues me too. I love that the italic actually changes the serifs! (I admittedly don't know the implications of that on an accessibility standpoint, but on an aesthetic one, I'm into it.) https://fonts.google.com/specimen/IBM+Plex+Mono

lazerwalker:
Closing this for now, will re-open when/if if I want to revisit!

PR #63, opened 2020-08-04
Update style.css

kawa-kitsuragi:
#25 Imported and used 'IBM Plex Mono', a monospace typeface I think is pretty nifty, including import of bold and italic options!

If we don't like how this looks, I can change it to a different one pretty easily, as long as it's on Google Fonts.


kawa-kitsuragi:
Seems like the check failed due to some sort of authentication issue.

lazerwalker:
Thanks for this!

On a technical level, I'd love to embed the font directly in this GitHub repo. The application bundle is currently completely standalone, so I'd love to maintain that by having everything we need to run the website living here instead of hitting Google's CDN directly. LMK if you need help with this!

I'm not particularly happy with how this looks right now, but I think it could also very easily be all of the other whitespace/layout issues I haven't cleaned up yet. My instinct is to merge this in as-is once it's loading locally instead of from a CDN, wait until there's generally been more love given to the visual layout, and then hopefully I'll either change my mind or it'll help clarify what specifically about the typeface is rubbing me the wrong way.

(screenshot for reference)
Screen Shot 2020-08-04 at 12 53 11 PM

lazerwalker:
Awesome! Thanks for figuring this out yourself, sorry I didn't pop back in to give direction.

PR #64, opened 2020-08-04
Issue 43, Left click also opens name menu

jim-adventureprogames:
Found this solution, and it seems like you used it as well for the clickable name at the top of the screen.


lazerwalker:
Awesome!

Hopefully this didn't take you a lot of research time, realizing I probably should have posted the solution once I figured it out for the top menu.

PR #65, opened 2020-08-04
Issue 19: Can type emojis via shortcuts. 

jim-adventureprogames:
Grabbed node-emoji and it seems to do the trick. Claims to be able to add custom emoji as well.

https://www.npmjs.com/package/node-emoji

I have never done this ridiculous thing where you just run install on some other code base and poof it works, this is all scary and new and different.

yabm_emojify

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ node_modules
 dist
 .cache
 .DS_Store
+.idea
 **/.DS_Store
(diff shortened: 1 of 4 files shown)


lazerwalker:
Looks good to me!

If we get to a point where we want to add custom emoji, we'll need to handle things on the server (and I realized I've already got code that handles that from a past project! Will drop a link in the relevant issue) but I think this is 💯 as the easiest way to get this in for now.

PR #67, opened 2020-08-05
Fix bug causing the profile to never update

MoyTW:
New users could not register their data and old users could not update their profile. Fixed bug by saving when update function called.

**Before:**

After submitting new data, it clears the form and shows `isRegistered` as empty:

Screenshot from 2020-08-05 14-08-52

**After - Registration**

When submitting it updates and lets you pass through:

workingupdate

**After - Updating Profile**

Showing previous & updated values before hitting Submit:

beforenewsubmission

Showing values were successfully saved when returning to Profile screen:

afternewsubmission

### DIFF
--- a/server/updateProfile/index.ts
+++ b/server/updateProfile/index.ts
@@ -28,6 +28,8 @@ const httpTrigger: AzureFunction = async function (
     data.twitterHandle = twitterHandle;
   }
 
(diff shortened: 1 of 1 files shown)


lazerwalker:
*facepalm* what an embarrassing regression. Might be time for me to think about testing 😂

Thanks for catching!

PR #74, opened 2020-08-08
Issue 68: Add /me and /emote commands

MoyTW:
Adds /me and /emote commands. They resolve identically:

emote

Also includes a fix for an error check in the /look command, which I didn't bother to try to reproduce.


lazerwalker:
Looks perfect! Thanks so much!

PR #92, opened 2020-08-12
Changes from running the linter

MoyTW:
Best do it in a independent PR so it doesn't create an unreadable diff when somebody runs it against their feature.


lazerwalker:
Awesome, thanks for this. I've been trying to do this, but having some weird issues with my local setup.

PR #96, opened 2020-08-13
Issue 79: mods can delete messages

MoyTW:
**User View**
Before:
Screenshot from 2020-08-13 15-04-00
After:
Screenshot from 2020-08-13 15-05-05

**Admin View**

Screenshot from 2020-08-13 15-02-58
Screenshot from 2020-08-13 15-04-19
Screenshot from 2020-08-13 15-04-36
Screenshot from 2020-08-13 15-04-54

**LocalStorage comparison - messageId is same across clients**

Screenshot from 2020-08-13 15-05-44
Screenshot from 2020-08-13 15-07-21

### DIFF
--- a/src/components/MessageView.tsx
+++ b/src/components/MessageView.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useContext } from 'react'
 import {
   Message,
   MessageType,
@@ -15,6 +15,11 @@ import {
   ModMessage
 } from '../message'
 import NameView from './NameView'
+import { MenuItem, ContextMenuTrigger, ContextMenu } from 'react-contextmenu'
+import ReactTooltip from 'react-tooltip'
+import { UserMapContext } from '../App'
+
+const removableMessageTypes = [MessageType.Chat, MessageType.Emote, MessageType.Shout]
 
 export default function MessageView (props: { message: Message; id: string }) {
   const { message } = props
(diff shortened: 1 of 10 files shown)


MoyTW:
I've used neither Azure Functions, nor React, nor, for that matter, Typescript before. In fact I've never used javascript. So, if there's any super jankiness in the code please point it out.

src/components/MessageView.tsx
```diff
@@ -42,6 +46,41 @@ export default function MessageView (props: { message: Message; id: string }) {
...
+
+  if (!playerIsMod) {
+    return (
+      <span>{props.message}</span>
+    )
+  } else {
+    return (
+      <span className="deleteMenu">
+        <ContextMenuTrigger id={props.messageId} renderTag="span" holdToDisplay={0}>
+          {props.message}
```
lazerwalker, line 68: Instead of passing in `props.message` as a property, it might make sense to use `props.children` here. That would let us do something like e.g.

`<DeleteMessageView id='1234' messageId='5678'>I am a <strong>message with nested tags!</strong></DeleteMessageView>` and have the nested React components properly rendered.

Feels a bit more React-y, and might be helpful with new message types? 

(https://reactgo.com/react-props-children/ if you haven't seen this yet)

src/components/MessageView.tsx
```diff
@@ -42,6 +46,41 @@ export default function MessageView (props: { message: Message; id: string }) {
...
 }
 
+const handleDeleteMessage = (e, data) => {
+  const doDelete = confirm(`Are you sure you would like to delete the message '${data.message}'?`)
+  if (doDelete) {
+    deleteMessage(data.messageId)
+  }
+}
+
+const DeleteMessageView = (props: { messageId: string, message: string, id: string }) => {
```
lazerwalker, line 56: Unless I'm missing something, you don't seem to actually use `id` here, yeah? I think we can delete it?

src/reducer.ts
```diff
@@ -302,6 +310,18 @@ export default (oldState: State, action: Action): State => {
   return state
 }
 
+
+
+function deleteMessage (state: State, messageId: String) {
+  const target = state.messages.find(m => isDeletable(m) && m.messageId === messageId )
```
lazerwalker, line 316: I don't yet 100% understand the TS issue you mention without diving into the code in an editor myself, but: since message IDs should be unique, we can probably shrink down to one `isDeletable()` call by removing the one from the `find` function?

MoyTW: Here's my understanding of what happens:

The `Message` class isn't required to have an `id` field; only Chat/Shout/Emote are. Therefore, you can't do `m.messageId === messageId` or even `m.messageId && m.messageId === messageId`. The `isDeletable` enforces the typing assurance of "the output of this function is in the class of Chat/Shout/Emote", which lets the `m.messageId === messageId` work.

However, the typing information doesn't get propagated out to `target`; target is of type `Message`. Therefore, we run into the same problem, except this time it's not on `messageId`, it's on `message` itself, since some message types (say, `MovedRoomMessage`) don't have a `message` field. So we need to again call the type guard function, so that `target.message = 'WORDS'` can be assured to be called on Chat/Shout/Emote.

You could solve this in a different way by explicitly casting the result of the find to `ChatMessage | ShoutMessage | EmoteMessage` but I thought it was better to just have it in the `isDeletable` function.

Issue #89, opened 2020-08-11
Replace the YouTube video in the Theatre with our actual Twitch channel


lazerwalker:
We're https://www.twitch.tv/roguelike_con

kawa-kitsuragi:
Closed per #98

Issue #88, opened 2020-08-11
Move to https://chat.roguelike.club or similar


lazerwalker:
Britta offered to add me to the DNS config. I should follow up on that.

lazerwalker:
I have access to our domain settings, but our DNS is hosted elsewhere. I emailed Douglas, who I believe manages this.

lazerwalker:
Appears to work!

PR #98, opened 2020-08-14
Update room.ts

kawa-kitsuragi:
Swap Youtube for Twitch in the embed for the theatre, for #89 .

Do note that with how Twitch embeds work, if we switch domains ( #88 ), we will need to change the 'parent' parameter in the embed. 

Twitch documentation on 'parent': https://discuss.dev.twitch.tv/t/twitch-embedded-player-updates-in-2020/23956


lazerwalker:
Ahh, thank you for also fixing my "sculls" typo, that's been driving my crazy for ages 😂

PR #99, opened 2020-08-14
Some styling for the Edit Profile view

annie:
Made the Edit Profile view a bit prettier 😄 

It might be fun to add some other things to this as well? Like an option to choose an alignment (chaotic good etc), or a pixel editor to create a lil avatar.

Screen Shot 2020-08-13 at 22 42 22

Also, I'm not sure what the expected behavior is for viewing a profile, but it's appearing in an unreadable transparent overlay for me. I didn't see an issue about this, but I can add one if that'd be helpful!

Screen Shot 2020-08-13 at 22 54 18

### DIFF
--- a/src/components/ProfileEditView.tsx
+++ b/src/components/ProfileEditView.tsx
@@ -2,6 +2,8 @@ import React, { useState } from 'react'
 import { updateProfile } from '../networking'
 import { PublicUser } from '../../server/src/user'
 
+import '../../style/profileEditView.css'
+
 interface Props {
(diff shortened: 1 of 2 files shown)


lazerwalker:
This looks awesome, thanks so much!!

Agree it would be super-fun to brainstorm about what other fields would be most fun to add. I'll make an issue for discussion :)

Thanks for pointing out the profile view layout issue too — if you want to make an issue, that would be super helpful!

PR #106, opened 2020-08-17
Notes

lazerwalker:


lazerwalker:
For anyone curious about what this currently looks like: 

Screen Shot 2020-08-16 at 10 56 01 PM

lazerwalker:
I thiiiink my axe-linter prefs might only go into effect when this merges into master.

If the next PR breaks on the same error, remove axe-linter from this project :(

lazerwalker:
(for context: this is a reasonable negative, but one that we're handling elsewhere — the key handler needs to be a global one rather than bound to the element with the click handler.)

PR #111, opened 2020-08-18
Save userMap to localstorage

MoyTW:
It saves on every state update, which I don't think is a huge deal, since I wouldn't expect more than...several hundred items, at most? However, we could also do a deep copy of the users map and compare it before saving for changes, if localstorage perf is a concern but I think it can totally take it.

beforerefresh

afterrefresh

Also it looks like I'm getting double disconnect messages; I'm not sure what's going on with that.

Note that you can forcibly reproduce the bug by manually wiping the userMap key out of your localstorage (though why would you do that?) - and it's null guarded such that if you do it doesn't explode.

backwardscompatible

localstorage after saving:

localstorage

### DIFF
--- a/src/reducer.ts
+++ b/src/reducer.ts
@@ -311,7 +311,9 @@ export default (oldState: State, action: Action): State => {
 
   if (action.type === ActionType.LoadMessageArchive) {
     state.messages = action.messages
-    state.userMap = action.userMap
+    if (action.userMap) {
+      state.userMap = action.userMap
+    }
   }
 
   // Notes
--- a/src/reducer.ts
+++ b/src/reducer.ts
@@ -78,17 +78,6 @@ export default (oldState: State, action: Action): State => {
   const state: State = JSON.parse(JSON.stringify(oldState))
   state.prepopulatedInput = undefined
 
-  // Only join local users into the userMap if we don't already have them - server users are fresher & may contain updates
-  const localUserMapString = localStorage.getItem('userMap')
-  if (localUserMapString) {
-    const localUserMap: { [userId: string]: MinimalUser } = JSON.parse(localUserMapString)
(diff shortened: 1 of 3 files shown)


src/reducer.ts
```diff
@@ -78,6 +78,17 @@ export default (oldState: State, action: Action): State => {
   const state: State = JSON.parse(JSON.stringify(oldState))
   state.prepopulatedInput = undefined
 
+  // Only join local users into the userMap if we don't already have them - server users are fresher & may contain updates
+  const localUserMapString = localStorage.getItem('userMap')
```
lazerwalker, line 82: We probably don't want to do this every reducer run, only on pageload, yeah?

My assumption was this would happen at the same time we load messages from localStorage. We could even use the same timestamp check, since if we don't have any old messages floating around we don't have use for old user map data.

src/reducer.ts
```diff
@@ -349,6 +360,10 @@ export default (oldState: State, action: Action): State => {
     }
   }
 
+  // The userMap is always saved so if somebody writes new functionality that edits the userMap, they don't have to remember
+  // to save it.
+  localStorage.setItem('userMap', JSON.stringify(state.userMap))
```
lazerwalker, line 365: I'm mildly in favor of still limiting this to the UserMapAction codepath?

I can anticipate other functionality that edits the userMap (see https://github.com/lazerwalker/azure-mud/issues/112), but I think it's unlikely there would be changes to the userMap functionality where it doesn't still make sense to dispatch a UserMapAction. And given that, it does seem wasteful to keep writing this every reduce.

MoyTW:
Updated MR as per comments

lazerwalker:
Perfect, thanks!

PR #118, opened 2020-08-20
Video presence

lazerwalker:


lazerwalker:
For anyone following along at home: yikes, this was a large set of changes that I thought went in already, but in the kerfuffle with our previous Azure Functions instance breaking, it never did.

This was a messy rebase, it wouldn't surprise me if we see some regressions here. This is what tests should be for :)

lazerwalker:
Merged locally and pushed so I could squash all this ESLint garbage into a single commit.

PR #142, opened 2020-08-24
Issue 141: Fix whisper click popup

MoyTW:
whisperfix


lazerwalker:
Huh, I wonder how long this has been broken! Thanks for catching + fixing.

PR #172, opened 2020-08-26
Show whisper chat in profile view

annie:
Show whisper chat messages in the user's profile view! ✨

Kapture 2020-08-26 at 17 57 41

A few notes:
- I didn't add incoming notifications yet, so that's still a TODO
- I kept the italicizing for whisper messages because it seemed more ~ whispery ~
- I wasn't sure what to do with the Whisper option in the user context menu, so I went with the somewhat inelegant option of mashing Profile and Whisper into Profile/Whisper (lol). Very open to better suggestions for that!
- Whisper chat view on mobile looks super wonky - for some reason, the font size of the chat messages gets huge. Gonna try to figure out what's going on, but I don't think that necessarily blocks this PR.

### DIFF
--- a/src/components/InputView.tsx
+++ b/src/components/InputView.tsx
@@ -5,7 +5,10 @@ import { SendMessageAction } from '../Actions'
 import '../../style/input.css'
 const emojifier = require('node-emoji')
 
-export default function InputView (props: { prepopulated?: string }) {
+export default function InputView (props: {
+  sendMessage: (message: string) => void,
+  prepopulated?: string
+}) {
   const dispatch = useContext(DispatchContext)
   const [input, setInput] = useState('')
 
@@ -16,17 +19,11 @@ export default function InputView (props: { prepopulated?: string }) {
 
   const checkEnter = (e) => {
     if (e.key === 'Enter') {
-      onClick()
+      props.sendMessage(input)
     }
   }
 
-  const onClick = () => {
-    dispatch(SendMessageAction(input))
-    setInput('')
-  }
-
   useEffect(() => {
-    console.log('useEffect')
     document.getElementById('chat-input').focus()
     if (props.prepopulated) {
       setInput(props.prepopulated)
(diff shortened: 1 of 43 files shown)


src/components/InputView.tsx
```diff
@@ -5,7 +5,10 @@ import { SendMessageAction } from '../Actions'
 import '../../style/input.css'
 const emojifier = require('node-emoji')
 
-export default function InputView (props: { prepopulated?: string }) {
+export default function InputView (props: {
+  sendMessage: (message: string) => void,
+  prepopulated?: string
+}) {
   const dispatch = useContext(DispatchContext)
```
lazerwalker, line 12: This is unused now, yeah?

lazerwalker:
Ahh, amazing! :D

My gut instinct is it makes sense to still have both "Profile" and "Whisper" as separate actions in the user menu? From the perspective of someone who doesn't understand what's going on yet, I suspect seeing two separate items probably makes more sense than one combined one, even if they ultimately do the same thing? I dunno.

One other thing that worries me: without notifications, I don't think anybody has any way to know if somebody's sent them a message? A few ways we could handle this that come to mind:

* Putting whispers back into the main chat view, so they're mirrored both there and in the side pane
* Whenever someone sends you a whisper, proactively open their profile pane. This could get weird if e.g. you're in the middle of sending a whisper to somebody and then a new whisper comes in.

My gut instinct is the first might make more sense for now since it has fewer edge cases? I'm sort of assuming this problem is temporary until we have a real notification system.

annie:
i think having separate Profile and Whisper options (that essentially do the same thing) makes sense for now! i can also make it so that Whisper autofocuses the chat input when the profile opens.

hmm yeah i see your point about notifications. agreed that putting them back in the main chat view seems easiest for now. i also might try making it so that clicking on the message opens the profile sidebar - that way they'll function as kind of in-chat notifications. i'll try it out and see how it feels!

annie:
Typescript has thwarted me lol. i couldn't figure out how to do the autofocus & opening profile when clicking Whisper message, but i think this is ok for now in the interests of getting something functional out for the meetup. lmk what you think!

lazerwalker:
Cool! Yeah, definitely good enough for now :).

Thanks so much!

PR #173, opened 2020-08-26
quick fixes for issues 149 + 164

kawa-kitsuragi:
no highlight boxes, fixed the default theme selector behavior


lazerwalker:
Thanks!

PR #174, opened 2020-08-27
Issue 13: write only audit log

MoyTW:
This doesn't fully close issue 13 - it only includes the audit log. This is sufficient for the meetup, since the meetup is only 3 hours and we'll have the ability to pull the audit logs directly out of Azure. However work still needs to be done to close out the issue.

auditlog

### DIFF
--- a/server/src/authenticate.ts
+++ b/server/src/authenticate.ts
@@ -34,10 +34,11 @@ export default async function authenticate (
   if (audit) {
     context.bindings.tableBinding = [{
       PartitionKey: user.id,
-      RowKey: Date.now().toString(), // I suppose with an appropriately zealous bot and a lot of luck, you could hit this timing window
+      RowKey: Date.now().toString(), // With an appropriately zealous bot and a lot of luck, you could hit this 1ms timing window, so it's not foolproof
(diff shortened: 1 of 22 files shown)


server/src/authenticate.ts
```diff
@@ -18,7 +19,29 @@ export default async function authenticate (
...
+  }
 
   const user = await getFullUser(userId)
-  return await handler(user)
+  const handled = await handler(user)
+
+  if (audit) {
+    context.bindings.tableBinding = [{
+      PartitionKey: user.id,
+      RowKey: Date.now().toString(), // With an appropriately zealous bot and a lot of luck, you could hit this 1ms timing window, so it's not foolproof
```
lazerwalker, line 37: I wonder if it's safer to just generate a UUID?

lazerwalker:
Cool! Looks great, other than potentially moving RowKey over to a UUID or something more guaranteed to be unique?

Planning on shifting today from the current production deployment to an Azure instance where I can add others as collaborators, so I'm not the sole point of access for these.

MoyTW:
Updated to use UUID:

uuid

lazerwalker:
Merged, but I haven't deployed the prod server yet

PR #179, opened 2020-08-28
Issue 173: slash command validation

MoyTW:
The command list could theoretically in the future be provided by the server, but for right now it just sits on the client. This means if we add any new slash commands we also need to go add them to this structure.

I plan to also implement the /help command, which will use this object to print out the various items commands. Once that's implemented I'll add "Refer to /help or the HELP button in the upper-left corner " (if that's where it ends up) "for more info."

Screenshot from 2020-08-28 14-44-11

### DIFF
--- a/src/SlashCommands.ts
+++ b/src/SlashCommands.ts
@@ -1,0 +1,24 @@
+interface SlashCommand {
+    name: string;
+    description: string;
+    invocations: string[];
+}
+
+const SlashCommand = (
+  name: string,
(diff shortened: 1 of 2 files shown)


MoyTW:
For last commit, Fix Whispers:

Screenshot from 2020-08-28 15-13-02

I haven't checked for conflicts against the big whispers overhaul; should probably do that.

MoyTW:
No conflicts.

PR #181, opened 2020-08-28
Add a map!

lazerwalker:
Screen Shot 2020-08-28 at 7 29 21 PM

Screen Shot 2020-08-28 at 1 02 11 PM


lazerwalker:
a11y warning is expected/okay — I think it's acceptable the map isn't accessible, since the sidebar exists. It would be _such_ a pain, not worth it for now.

PR #187, opened 2020-08-29
Issue 150 & Issue 162

apepers:
- Add a 'same room' message that shows up when you click a room you're already in, and make it so no add/remove player actually happens (so other players shouldn't see anything)
- If you try to add a post it that is blank, it won't add anything (including the case of cancelling the modal)


kawa-kitsuragi:
Also updated the colour for the captions so that they're visible on all the CSS styles, and point to the right caption URL for the event

lazerwalker:
Both look good, thanks!

This will also be a good test for whether my new CI pipeline to automatically deploy the server works :)

Issue #208, opened 2020-08-31
Whisper panel should never show a horizontal scrollbar


lazerwalker:
It was for me, despite messages fitting on-screen. Just `overflow-x: hidden`, presumably.

PR #219, opened 2020-09-01
Moderator changes, restores user counts in sidebar, schedule change

### LINKED ISSUE #196: Distinguish moderators with accessible formatting
White names vs pink names may be hard to see for people with low vision or colorblindness, so let's distinguish mod names with something more visible! No specific opinions on implementation.

### LINKED ISSUE #213: Return room player counts to the left sidebar

kawa-kitsuragi:
Closes #196 
* Moderators now have both '[Moderator] ' before their username and are highlighted in a high-contrast way versus normal users
In 'Default' theme, the moderator has a gray background.
In 'Solarized Dark' theme, the moderator has a pale background, the background of 'Solarized Light'.
The colors all flip for the 'Solarized Light' theme.

* This is also reflected in Code of Conduct text.

Closes #213
* Restore user counts on sidebar

Also updated the schedule to not be that of the preview - but will need to be fixed again when we determine speaker order


kawa-kitsuragi:
Oh yeah, also fixes #208.

The whisper panel doesn't have a scroll bar anymore.

PR #239, opened 2020-09-22
Issue 93: move spam setting

MoyTW:
VERSION 2! Now, it has:
+ Server settings for a list of rooms where you never see the messages
+ Server settings for the number of people before it switches to hide messages mode
+ Update window for server settings (mod only)
+ Local setting for showing all movement messages

Screenshot from 2020-09-29 20-03-38
Screenshot from 2020-09-29 19-50-33
Screenshot from 2020-09-29 19-51-05
Screenshot from 2020-09-29 19-51-21
Screenshot from 2020-09-29 19-51-30
Screenshot from 2020-09-29 20-00-58
Screenshot from 2020-09-29 20-01-07
Screenshot from 2020-09-29 20-01-47
Screenshot from 2020-09-29 20-02-16


style/profileEditView.css
```diff
@@ -30,14 +30,15 @@ input[type=text] {
 }
 
 .form {
-  width: 780px;
+  width: 600px;
```
lazerwalker, line 33: Without pulling down your changes to test locally: are you sure this doesn't break the styling of the profile edit view? 

(If yes, probably makes sense to separate the two forms out into different CSS classes)

src/components/SettingsView.tsx
```diff
@@ -45,50 +45,60 @@ export default function SettingsView () {
...
         <label htmlFor="themeSelectionForm">Select Theme:</label>
         <div className='radio'>
-          <input type = "radio"
-            id = "theme"
-            value = "default"
-            checked = {selectedTheme === 'default'}
-            onChange = {handleThemeSelection}
-          />
-          <label>Default (Dark)</label>
+          <label>
```
lazerwalker, line 48: Looks like these were slightly incorrect from an accessibility standpoint before your changes!

So, each of these radio inputs should have their own id (e.g. `id='theme-default'` instead of them all being `id='theme'`).

From there, each `label` should be a sibling element of the `input` (rather than a parent/wrapper), but with a `for` element that refers to the input's id.

e.g. this example should be:
```html
<input type = "radio" value = "default" id = "theme-default" checked = {selectedTheme === 'default'} onChange = {handleThemeSelection} />
<label for='theme-default'>Default (Dark)</label>
```

lazerwalker:
Mostly looks good!

Other than the few comments I left, two design thoughts:

1. We know that people don't want move messages in the theatre. I feel like having messages hidden should be the default behavior.

2. I think it would be more intuitive if the checkbox was "Always show movement messages" instead of "hide movement messages". Having a checked-off checkbox represent "yes, show me more things" makes more sense to me, and given #1 this also means that the default state would be unchecked instead of checked, which also feels cleaner.

3. I don't think we should expose the person count variable to players. Just trying to explain that feels like a lot of work, and I feel like finding the optimal value there is our job as designers.

style/profileEditView.css
MoyTW: Well, it looked...fine? but I don't know how CSS or layouts work. It's mostly that if it's 780 it gives me a horizontal scroll bar on my laptop for no discernible reason.

src/components/SettingsView.tsx
MoyTW: ...huh. Okay. Is it fine that it passes the linter now, or is the linter too generous?

MoyTW:
For point 1, sure, can do.

For point 2 - "Always show/Never show" is different from the "Hide all/Hide some" which we have. We have 3 states:
a) Show all messages
b) Show messages when it's not very busy
c) Show no messages
and the current toggle toggles between [b, c] whereas I think the proposed toggle would toggle between [a, c]

I mean I could just change it from a checkbox into a 3-way checkbox thing like with the styles?

For point 3 - what would be a good number? I have 6 right now. Or do we want to just drop it for a binary all/none?

src/components/SettingsView.tsx
lazerwalker: I think we're fine for now. This feels unidiomatic to me, but after some research it does still produce the correct screen reader behavior.

lazerwalker:
Ahh, interesting point about a/b/c.

Thinking about this more: I suspect almost nobody wants to see enter/leave messages in the theatre. I suspect most people will gain value from join/leave messages in smaller rooms. I'm having a hard time thinking of a use case where either of those will not be true.

This is complicated enough that I think you're right we'd want a 3-way option if we wanted to give people a choice. I increasingly think we just don't want to give people a choice. People who either don't want to see leave/join messages in small rooms, or do want to see them in the theatre, sound like a pretty bad edge case. The only likely design failure we'd be looking at is if we just happen to get the threshold wrong.

MoyTW:
I think there's a case for "wait, Alice isn't here anymore? *scrolls back up* when did Alice leave, did she see that last message(s)?". A simple "show all/show some" toggle would be fine though.

I'm not sure how to set the threshold though.

lazerwalker:
Yeah, I think that makes sense.

For the threshold: My gut instinct is something like 20? I realistically think what we're designing for is "the theatre vs everything else", so erring on the larger side is probably okay/good.

MoyTW:
My reasonably strong threshold for transitioning between "small group" and "large group" modes is 8. See also my proposal to cap video chat at 8 here #284

lazerwalker:
So, I think the UX expectations here are different from hard capping videochat.

With videochat, erring on a smaller number makes sense for a number of reasons, not the least of which is that we can be very explicit about "sorry, chat is full".

With this, if we do things correctly, people aren't going to consciously notice we've done anything. So the question becomes, if we get the number wrong and people notice, what happens in that fail state.

I think we're better off erring too high, in which case people are slightly annoyed by visual spam, than we are by erring too low, in which case people get confused because they can't tell when somebody leaves.

MoyTW:
`aren't going to consciously notice we've done anything` hmm, I never really thought that was something that was possible? I guess my implicit assumption was that users would definitely notice the presence/lack of movement messages, because it's...pretty hard not to, if any room has them at all.

My sort of thesis is that more than half of all messages in any given feed are going to be movement messages, even in the highest conversation areas, just because:
1. moving is way easier than talking
2. by volume, there will be a lot of people, and one person entering and leaving takes up 2 lines. A throughput of even 1 person every 20 seconds makes enough noise to actually match the speed of many text conversations
3. until they are explicitly drawn into a conversation, people have basically zero stickiness for an individual room, so they'll just sort of hop around looking for something cool

The other thing is, if we only have a "Show All/Show Some" the user can't fix it to show *less*, they can only fix it to show more, so we should by default err on less.

kawa-kitsuragi:
Thinking of this, perhaps above the threshold a message that comes across that just says something along the lines of "Many people are entering/leaving the [room]", to give a sense of the motion that something is happening and not be abruptly "you can tell who is entering/leaving in room X and now in room Y you can't"? Not sure how that will be to implement!

lazerwalker:
I absolutely think people won't notice this if our design is correct. People will notice the messages, or lack thereof, if they're disruptive. If they're not, I'm sure you could point out to someone "hey, have you noticed how sometimes the messages show and sometimes they don't?" and they'd recognize it, but design choices like this otherwise tend to fall into the background if they're working the way people sort of intuitively expect them to. 

I think a lot about how e.g. Slack's logic about whether to send a notification is unbelievably complicated, but as a user the mental model is purely "I should get a notification whenever it makes sense": https://twitter.com/codevisuals/status/838881724016787457/photo/1

I'm not sure I agree with your thesis, but it's hard to come up with a solid argument that isn't "neither of us has concrete data, but my back-of-the-envelope math is different from yours about what the hypothetical rate of conversation or movement will be".

Another thing that's crossed my mind: there were times in the preview when I felt slightly awkward about leaving a room because of the leave message. Which I think might suggest erring on the side of showing fewer leave messages, but I'm not exactly sure.

lazerwalker:
@kawa-kitsuragi I think there's one use case there that would be potentially helpful: if I enter a room and then you immediately enter the room before anyone says anything, those two enter messages could theoretically be collapsed into a single "lazerwalker and kawa enter the room" message. I think trying to collapse join/leave messages in any other situation is tricky from a design perspective. Haven't thought through how technically viable this would be to happen on our timescale -- we don't yet have any instances of a message actively being altered after the fact, outside of mods having the ability to delete inappropriate messages.

lazerwalker:
Since we have differing opinions and so little data: I wonder if it makes sense to store the limit in Redis, and we can raise/lower it based on actually seeing how people behave during the conf?

MoyTW:
> @kawa-kitsuragi I think there's one use case there that would be potentially helpful: if I enter a room and then you immediately enter the room before anyone says anything, those two enter messages could theoretically be collapsed into a single "lazerwalker and kawa enter the room" message. I think trying to collapse join/leave messages in any other situation is tricky from a design perspective. Haven't thought through how technically viable this would be to happen on our timescale -- we don't yet have any instances of a message actively being altered after the fact, outside of mods having the ability to delete inappropriate messages.

I tried doing that but it was really hard to get it to look half-decent and I gave up. But it's doable! You lose a lot of data with timestamps and stuff but it works.

MoyTW:
For now I'll look into storing the value on the server.

MoyTW:
Uh, hold on. It looks like it did a bad merge on the Room - still has all the showcases. Ugh.

lazerwalker:
🎉

PR #248, opened 2020-09-23
Showcase: gesuido & skeletons for other games

MoyTW:
Adds Gesuido draft (frontends, what are they even) & skeletons for rest of the hall

Also caps the height of each individual entry to 600 & adds scrolling if you shrink the page. The image shrinking with Gesuido is jank because I don't do frontend stuff and it's 100% me + Google trying random things. 

Screenshot from 2020-09-23 15-46-07

Screenshot from 2020-09-23 15-46-27

Screenshot from 2020-09-23 15-46-47

Screenshot from 2020-09-23 15-46-56


MoyTW:
If anybody knows how to do CSS for Gesuido that puts the text/images horizontally until the width gets small and then stacks them vertically that would be really helpful, ha.

PR #249, opened 2020-09-23
WIP: Wrangling Rift Wizard

MoyTW:
Hmm. This is vexing! I'm vexed.

I wonder if GitHub does the same `WIP:` detection that GitLab does. Anyways. I seem to have inadvertently committed myself to doing frontend web design for the Games Showcase. Here be the extremely generous, mind, spec:

```
Screenshots attached.  They tend to look bad when scaled- if we need to make them smaller can we either do it by exactly 50%, or can we crop rather than scale?

Where to buy: https://store.steampowered.com/app/1271280/Rift_Wizard/.
If possible lets use this pretty html widget:
<iframe src="https://store.steampowered.com/widget/1271280/" frameborder="0" width="646" height="190"></iframe>	<br><br>
```

The screenshots are huge, 1600x900. So, there's no way we can possibly fit those. The iframe width is 646; I tried resizing by cutting the width and upping the height but that just cut off the text on the bottom.

I tried hacking something together, but am not super happy with it. I would appreciate input from somebody who has design or frontend experience. I put the images at width=400 & with the widget, here's what it looks like:

Browser fullscreen:

Screenshot from 2020-09-23 16-21-04

Browser on my computer - half screen:

Screenshot from 2020-09-23 16-23-37

For comparison, half screen on a page that doesn't have the widget scales down better:

Screenshot from 2020-09-23 16-25-24


MoyTW:
So keeping the widget at `646` means the screen gets stretched out and given that our mobile is expected to be, uh, less than that it's not great. I have no idea how to test it on mobile without just pushing it on master and pulling it back out later if it doesn't work - @lazerwalker is that what you've been doing for the things you tested?

MoyTW:
Here's what it ends up looking like with the CSS changes from #256 in effect at a window width of ~930:
Screenshot from 2020-09-24 14-54-51
I think that's fine enough? The main question is should we drop the widget and just say "nope, our design limit is [x width]".

MoyTW:
I'm on the verge of just saying that Games Showcase entries have a minimum [650, 600] box to play with, and if they overflow it overflows vertically. Showcase folks keep asking me "how big should the screenshots be" so that seems reasonable enough.

PR #256, opened 2020-09-24
Make padding, gap, right margin scale

MoyTW:
Mostly this is an issue when you view at lower resolutions - have 200px
on the right side + 50px width & padding is a LOT of wasted space when
you downscale your browser.

With fixed margins & padding & right column:

Screenshot from 2020-09-24 14-45-45
Screenshot from 2020-09-24 14-46-03

With variable margins & padding & right column:

Screenshot from 2020-09-24 14-47-40
Screenshot from 2020-09-24 14-47-59

### DIFF
--- a/style/style.css
+++ b/style/style.css
@@ -47,17 +47,17 @@ body {
 }
 
 #app {
-  padding: 0 50px;
+  padding: 0 3%;
   display: grid;
-  grid-template-columns: 250px 1fr 200px;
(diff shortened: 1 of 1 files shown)


MoyTW:
You can see that it's still wasting a lot of right-side space inappropriately when you have a smaller window but it's less space at least! If I knew how to have it "eat" the right-side margin first as you scaled down that would be ideal.

PR #257, opened 2020-09-25
WizardWars info

quantumpotato:
hello, I added info about WizardWars.
Can images be shown here? We have an image: https://raw.githubusercontent.com/WizardWarsIO/wiz-server/master/screenshot.png

### DIFF
--- a/server/src/room.ts
+++ b/server/src/room.ts
@@ -92,7 +92,10 @@ export const roomData: { [name: string]: Room } = {
     description: `
       A number of booths line the sides of the hall.<br/><br/>
       <img src="images/puppy.jpg" alt="puppy" />
-      Here you can see Gesuido, by Ryosuke Mihara, Project Earth by Michael Taylor, WizardWarsIO by quantompotato, and Project Shiro by Garbriel Martinez.<br/><br/>
+      Here you can see Gesuido, by Ryosuke Mihara, Project Earth by Michael Taylor, WizardWarsIO by R. Riley Holmes and quantumpotato, and Project Shiro by Garbriel Martinez.<br/><br/>
+      <br/><br/>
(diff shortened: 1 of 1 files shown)


lazerwalker:
Thanks for this! 

Two options for images:

1. Host the image yourself and link to it directly. I wouldn't recommend linking to a GitHub CDN link, but enabling GitHub Pages and linking to that would work

2. Add the image to this repo. Anything dropped into the `images` folder in the root directory can be accessed easily via e.g. ``

lazerwalker:
(Do you think you might be able to also clean up the merge conflict? Sorry we're moving things around pretty rapidly right now as we're setting up the arcade space — I can pull down your changes and manually merge them if need be, but would be easier if that's something you can do :) )

quantumpotato:
Hey I'm having to do client work tonight but I can look at this tomorrow
night

MoyTW:
Okay, so, there are two things I think are possible:
1. We put the copy & screenshots & stuff up, like with everything else in the showcase - see the other entries for examples - and have a link to the game. This is easy, and fast.
2. We embed the game - it's "close enough" to our size limits for me to say okay, sure, we can go over a little. I got a prototype looking like this:
Screenshot from 2020-09-24 19-58-57
the only issue is I couldn't get the up/down keys to stop the app from scrolling the app as you play. Which is a blocker I think, but if somebody has a good idea how to stop that I'd be down to embed it. Unfortunately I do not actually have the expertise or time to fix that. If we do embed it I'd say we should *not* add any additional text, because it's already over the 600px size for showcases.

quantumpotato:
We tried disabling keyoverrides and it caused some other issues. Let's just
link to the site. http://206.189.227.251:8000 - and we can put a screenshot
in place of the embed?
Please use WizardWars.IO by R. Riley Holmes and quantumpotato

quantumpotato:
Closing in favor of https://github.com/lazerwalker/azure-mud/pull/268

PR #261, opened 2020-09-25
Add the ability to close off the space to non-mods [Closes #198]

### LINKED ISSUE #198: A way to 'close the doors' while the space isn't open
I'm imagining a flag (that could require a deploy, a UI menu option for mods, or a manual Redis change) that closes off the space to everyone who isn't a mod.

If the space is 'closed', and you log in as a not-mod, instead of the space you have a "The doors are closed" message.

Useful for the dead period between now and October, and more relevantly for the day-of the conference before we're officially open.

Bonus points if enabling it actively kicks everyone out of the space, so we can tell people to go home.

lazerwalker:
Screen Shot 2020-09-25 at 00 36 37


lazerwalker:
Opening as PR for visibility. See #198 for more info.

PR #285, opened 2020-09-29
A Dumb Thing

MoyTW:
The color changing is pretty cool. It's mostly the flavor text that's dumb.

And the CSS implementation. Thank you emacs for making this possible.

Screenshot from 2020-09-28 23-34-00
Screenshot from 2020-09-28 23-34-09
Screenshot from 2020-09-28 23-34-24
Screenshot from 2020-09-28 23-34-32
Screenshot from 2020-09-28 23-34-39
Screenshot from 2020-09-28 23-43-39
Screenshot from 2020-09-29 00-21-59
Screenshot from 2020-09-28 23-44-07


MoyTW:
Somebody who knows CSS tell me how stupid/genius/normal that CSS hack I put in for all the colors is.

kawa-kitsuragi:
I'm worried about how this will look on the other themes. We should weed out colors that won't stand out well against the backgrounds of Solarized Dark (hex `#002b36`) and Solarized Light (hex `#fdf6e3`.) 

Solarized absolutely supports 8 colors that work on both backgrounds, list here under 'The values': https://ethanschoonover.com/solarized/ More than those will probably require serious testing on all 3 backgrounds that I don't think we have time for.

Our style https://github.com/lazerwalker/azure-mud/blob/main/style/style.css uses the Solarized-approved versions of green, pink, and blue (and then reasonably close variants for the default black background) to represent links (green), warning text (pink), and the thumbs up upvote (blue). 

So what might be worth doing is defining `--yellow`, `--orange`, etc as CSS vars in the 3 sections of styles.css (`root`, `.solarized-dark`, and `.solarized-light`) and changing the function to grab the defined var.

lazerwalker:
+1, I think this is a really fun idea, but making sure it plays nicely with our color scheme seems like an important accessibility concern.

apepers:
It should actually be really easy to support the themes, with how you set it up. Like Kawa said, in the :root, .solarized-dark, and .solarized-light, you can define more variables for colours, like --orange, and define it so that in each theme it's a visible colour. Those solarized colours Kawa linked save you some work because you can use the same value across light and dark, but there's nothing actually stopping you from adding a colour not on that list, you just need to make sure you pick a hex value for it in light that is visible on light, and a different hex value for the same general colour in dark that works for dark.

Then all you have to do is pair your ValidColors list down to just however many variables you've defined for colors, you can leave it the same where it's like Orange = 'Orange'.

Then your css just has 
`.name.Orange` 
{ 
color: --orange 
}`

MoyTW:
So, really I appreciate all the help on colors there, but I also have literally never used CSS before and have no idea what any of that means. It sounds like I'd have to test every color (?) manually to make sure it works with each theme, and then use `--color: #hexvalue` to something with enough contract on each theme to make it compatible? Which, if I'm understanding correctly, sounds like faaaaar too much work to be plausible.

Looking at the Solarized website it seems like if I just limited the valid colors to [yellow, orange, red, magenta, violet, blue, cyan green] it will just work out of the box? That's actually very easy, it's just deleting a lot of stuff.

MoyTW:
...I thiiink I get it? I'll put up a commit shortly.

lazerwalker:
Coool, I suspect we'll eventually want to move this someplace other than Table A, but 👍 for now!

The merge conflict looks really simple, but GitHub's web UI isn't letting me do it, so I'll leave that to you to merge whenever :)

lazerwalker:
oh, lol, JK, I got a failure error message but the merge still went through

PR #286, opened 2020-09-30
More robust map

lazerwalker:
This does a lot of things!

1. Refactors the map to let us draw an ASCII map in an ASCII art editor
2. Adds a handful new rooms, and updates descriptions appropriately
3. Adds our first instances of sub-rooms: the kitchen has tables, and the showcase hall is 4 different rooms
4. Replaces the list of rooms sidebar with a small static mini-map
5. Focuses the map (both regular + mini-map) on the user's location if it's too small
6. Refactors the map modal so it tries to be large enough for the map
7. Adds a modal 'list of rooms' for accessibility purposes, since the map almost certainly isn't accessible

Screen Shot 2020-09-29 at 9 29 08 PM
Screen Shot 2020-09-29 at 9 29 04 PM


lazerwalker:
Accessibility failures are fine. I should probably add ignore comments to the file itself, but it's expected the map is not screenreader-friendly. This is why there's a plain list view provided, and the link to that list view appears earlier in the aria hierarchy than the weird invisible clickable areas.

PR #289, opened 2020-09-30
Easter egg commands, music in the dance floor

apepers:
Added music to the dance floor, which is good to go.
The major work is adding new commands for more playful interactions. The dance command is working properly (though my attempts to put it in its own file failed, I couldn't figure out how to call the dance function without hitting a 500 error).
The get/interact function had a few more to-dos left so I'm okay with holding off on the merge until those are done, but I wanted to show the direction I've taken so that if anyone *does* have suggestions of a tidier way to put it into its own file, I'm all ears.


src/config.ts
```diff
@@ -1,5 +1,5 @@
 // TODO: Eventually abstract this out of this repo
 // My .env trick I've used in other projects needs some touching up of its regex
 export default {
-  SERVER_HOSTNAME: 'https://roguelikecelebration-mud.azurewebsites.net'
+  SERVER_HOSTNAME: 'https://roguelikecelebration.azurewebsites.net'
```
lazerwalker, line 4: This is your personal dev backend, yeah?

server/sendChatMessage/index.ts
```diff
@@ -7,6 +7,59 @@ import authenticate from '../src/authenticate'
 import { getUserIdForUsername } from '../src/user'
 import { MAX_MESSAGE_LENGTH } from '../src/config'
 
+const danceList = [
```
lazerwalker, line 10: It would probably be nice to extract these lists out into an external TS file that's imported, this is pretty noisy for a file that's mostly business logic about message sending.

lazerwalker (commented):
Awesome, mostly looks good! Two really quick things.

Also, to confirm my assumption/understanding: I'm guessing the plan is to eventually add room descriptions for some rooms that suggests there are fortune cookies / potions?

src/config.ts
apepers: Yup sorry, meant to not include this one in the CL, I'll get it taken out

server/sendChatMessage/index.ts
apepers: Yeah - I tried putting the whole 'dance' function into an external TS file and exporting the function, but was running into 500 server errors. I was trying to base it off shout.ts but I had some confusion around whether or it it's awaitable and how to return properly, since unlike shout or whisper it isn't doing any other operations. If you could give me tips on how to move the whole function into an external file I'd love to do that - otherwise you're right I could at least just export the arrays.

apepers:
That's right - the plan is to have a basket of fortune cookies in the kitchen, and potions in the bar. I'm hoping to do some kind of italic or text distinction to give a hint that they're interactable, just in the description by hand. 

The idea is then to also in the check for interactable, if you get a match on cookie, also just check user room = kitchen, as the simple way to ensure you only interact with what's in your room. Given time a more systems driven option would be better... but for now I'm doing it Nethack style, if statements for days ^_^;;; 

(It is because of that though, to my point above about separate files, I'd really love if you do know how I could move the interact logic into a function I import and call since I hate to bloat index.ts and further and this one is gonna have even more branching logic

server/src/dance.ts
```diff
@@ -0,0 +1,32 @@
...
+  'starts doing the Yak dance (it’s real, look it up!)',
+  'moves their arms in a demonstration of ‘Big Fish, Little Fish, Cardboard Box’',
+  'stomps their feet in some sort of jacked-up footwork… Ubi folk dancing?',
+  'pulls a leek out of *somewhere* and starts spinning it around',
+  'finds a couple glowsticks and starts waving them to the beat',
+  'nods their head to the beat. Feels good.',
+  'pulls off some stair dancing technique - good at evading monsters, but a little crass, don\'t you think?'
+]
+
+export function dance (user: User, messageId: string, context: Context) {
```
apepers, line 19: @lazerwalker I actually did check in my external file, partly so I could ask for help on what about it is going wrong and how I could get it working, if you have the time!

server/sendChatMessage/index.ts
```diff
@@ -62,6 +115,52 @@ const httpTrigger: AzureFunction = async function (
...
+    // Todo, limit these to the right room
+    if (interactMatch) {
+      var inspectedObject = interactMatch[2]
+      if (inspectedObject.includes('cookie') || inspectedObject.includes('fortune')) {
+        // Inspecting a fortune cookie
+        context.bindings.signalRMessages = [
+          {
+            groupName: user.roomId,
+            target: 'emote',
+            arguments: [req.body.id, user.id, ('opens a fortune cookie that reads: ' + fortuneList[Math.floor(Math.random() * fortuneList.length)])]
```
apepers, line 141: @lazerwalker another question, if you have the time. Kawa had included some fortune cookies with links in them to previous talks, the code of conduct, etc. They wrote on the text for the fortune as HTML with the hyperlink - but the way this gets rendered in the end it escapes the characters and instead of a hyperlink you can text with the unicode mess and the href etc. Looking into options if I did want to try to render the text as HTML it all looks pretty messy, like using "dangerouslySetInnerHTML" - do you know of any way you'd be happy with? Low priority, since we don't *need* links in fortunes.

lazerwalker: We're currently using dangerouslySetInnerHTML for rendering the room descriptions (plus some manual parsing for the Twine-style links), I think it's fine if we use it for this case. 

It's mostly a vulnerability if you're using it to render user-generated text, but as long as we're only using it for content that's hardcoded into our server code it should be safe.

lazerwalker:
I think I understand what you're asking re: imports! Gonna word vomit for a bit here, but if this isn't what you're asking or if you have other stuff, feel free to ping me on Discord via DM or the dev server, more than happy to chat :)

What's confusing is there are a handful of different ways you can structure imports that all require slightly different JS/TS `import` syntax. For what we're doing, the differences don't matter, other than just making sure that the syntax of your import matches the syntax of the thing you're importing.

https://github.com/lazerwalker/azure-mud/blob/main/server/src/look.ts is a good example of a really short function that's importable like so: https://github.com/lazerwalker/azure-mud/blob/main/server/sendChatMessage/index.ts#L5. If you `export` a named variable or function (like `look` in that file), you can import it as `import { theNameOfTheExportedThing } from './pathToFile'`.

If you want to be able to `await` a function, the function just needs to have the `async` keyword in its declaration (like `look.ts`). If you want to call `await` from within a function, it itself needs to be `async`. There's some other stuff that applies with TypeScript specifically (namely, if you declare an explicit return value from an async function in the method signature, it needs to be a Promise instead of a raw value), but I suspect you can get away with not worrying about that.

server/sendChatMessage/index.ts
apepers: Okay awesome - I'll try that out tonight!

apepers:
For visibility, my to-do list left on this is:
- Make fortune cookie message only show up for user reading it, other people see 'cracks open a fortune cookie'
- Hook up polymorph potion using Em's item pickup system
- Hook up potion of cancelation using the same
- Add reference to fortune cookies to kitchen, with hyperlink using Em's system
- Add reference to potions in bar, with hyperlinks
- Add blessed and cursed fortune cookies

PR #292, opened 2020-09-30
While we do strive to be welcoming...

MoyTW:
...not, like, that welcoming.


MoyTW:
Kinda surprised typescript didn't throw a fit? That seems like the kinda thing it'd throw a fit for.

PR #299, opened 2020-10-01
Issue 293: happening now

MoyTW:
Okay, it didn't really work in the sidebar. I turned it into a button on the same level as map. Unsure if that was the best idea.

There are 2 parts:
1. It dynamically pulls from the schedule and tells you what rooms are associated with the event if any
2. It pulls from manual updates in ServerSettings to form sort of an ad-hoc announcement system, I guess?

You can edit these through a form in the ServerSettings. The whole thing is very Programmer.

Screenshot from 2020-09-30 22-15-34
Screenshot from 2020-09-30 22-09-02
Screenshot from 2020-09-30 22-09-29
Screenshot from 2020-09-30 22-10-08
Screenshot from 2020-09-30 22-10-41
Screenshot from 2020-09-30 22-11-16
Screenshot from 2020-09-30 22-12-11
Screenshot from 2020-09-30 22-12-25
Screenshot from 2020-09-30 22-12-53
Screenshot from 2020-09-30 22-14-28
Screenshot from 2020-09-30 22-14-50
Screenshot from 2020-09-30 22-15-02

### DIFF
--- a/server/src/types.ts
+++ b/server/src/types.ts
@@ -2,14 +2,13 @@ import { Room } from './rooms'
 import { User, MinimalUser, PublicUser } from './user'
 import { RoomNote } from './roomNote'
 
-
 export interface HappeningNowEntry {
   text: string,
   roomId?: string,
   externalLink?: string
 }
 
-function isHappeningNowEntry(obj: any): obj is HappeningNowEntry {
+function isHappeningNowEntry (obj: any): obj is HappeningNowEntry {
   return obj.text && (typeof obj.text === 'string' || obj.text instanceof String) &&
     obj.roomId ? (typeof obj.roomId === 'string' || obj.roomId instanceof String) : true &&
     obj.externalLink ? (typeof obj.externalLink === 'string' || obj.externalLink instanceof String) : true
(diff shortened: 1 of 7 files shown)


server/src/types.ts
```diff
@@ -2,25 +2,41 @@ import { Room } from './rooms'
 import { User, MinimalUser, PublicUser } from './user'
 import { RoomNote } from './roomNote'
 
+export interface HappeningNowEntry {
+  text: string,
+  roomId?: string,
+  externalLink?: string
+}
+
+function isHappeningNowEntry (obj: any): obj is HappeningNowEntry {
```
MoyTW, line 11: I'm having this awkward slapfight with Typescript with regards to server-side validation that makes me feel like a right idiot. This doesn't do what I really want it to do but it's Good Enough I think...but there's something I'm completely not understanding.

MoyTW:
With regards to the non-functional link element - I *do* want it to look and act like a link (be blue, when you mouse over it the cursor changes) - is there a way of styling a paragraph to do that, or something? And would that be...better, or worse, from an accessibility standpoint?

lazerwalker:
re: links, you can make them buttons with the CSS class `link-styled-button`

lazerwalker:
Can we rename the "Live from Here" to "Other Live Activities" or something like that? That phrasing doesn't quite sit right with me.

server/src/types.ts
lazerwalker: Yeah, I think there are ways this could be made a lot cleaner and do more what you're intending it to, but I think those changes also bubble up to refactoring how `toServerSettings` works. Happy to go back after the event and refactor stuff together if you're interested in learning/figuring stuff out, but also agree this is probably Good Enough for T-36 hours.

MoyTW:
Rename header & buttons styled as links

Screenshot from 2020-10-01 16-15-42

PR #303, opened 2020-10-02
Issue 194: limit message word length to 60

MoyTW:
Screenshot from 2020-10-01 19-13-35


lazerwalker:
Calling out that there's a small chance we'll run into someone who wants to legitimately post a longer-than-60-char URL. I still think this is the right choice for now — I think "just use a URL shortener" is sub-par but fine.

PR #305, opened 2020-10-02
Add links to match the visual map

MoyTW:
I'm...concerned about the level of interconnectivity because each of the links has a number that's gonna update on movement. We could relatively easily just erase a bunch of the connections? @lazerwalker mentioned you were seeing some sort of perf issue with the room links & updating too often, and after this MR the kitchen is a monster of a room, have 9 (!) connections.


lazerwalker:
Thanks for this!

I'm not sure we'll actually see a perf issue with the room links updating too often, but if we do, it won't be related to how many connections a specific room has.

tl;dr: React automatically re-renders a component when any of the pieces of data that serve as inputs to it (the props, or anything pulled in via a useContext) change. Now that we're pulling in all of roomData as a prerequisite of RoomView, instead of merely the data for the single room we care about, any time the global presence indicator changes we're re-rendering the RoomView. 

My instinct is still to leave it as-is for now, and if people moving around a lot slows down rendering for everyone, then we kill the presence indicators in the links for every room.

PR #308, opened 2020-10-02
Another Dumb Thing

MoyTW:
It is a series of rooms, accessed by the lounge. They're all hidden. You get a couple of choices in the middle, some of which can 'kill' you (send you back to the lounge). Once you get to the end you get an exhaustive list of every room and their ids so if you want to teleport to hidden rooms you can now. Presumably when items goes in tomorrow I might add, like, a fancy wizard hat at the end or something.

Honestly you should never let me write anything ever, especially not when tired.

This is what the whole thing is like.
Screenshot from 2020-10-01 23-28-10
Beware the bowties!
Screenshot from 2020-10-01 23-32-37
You start in the lounge and have a couple of normal rooms first though
Screenshot from 2020-10-01 23-27-31
Screenshot from 2020-10-01 23-27-41
Last room's a portal with a book that brings up the following modal
Screenshot from 2020-10-01 23-27-19
Here is the description & list of stuff
Screenshot from 2020-10-01 23-26-54
Screenshot from 2020-10-01 23-27-09


lazerwalker:
Merging this in as-is for now, but I feel like the list of all rooms you can teleport to is a bit overwhelming?

My instinct would be to (a) remove the public rooms from the explicit list, and (b) rename some of the secret rooms' roomIds? I feel like telling someone to teleport to the `hiddenClosetShoePathEnd` feels much more like exposing programming implementation details than giving people access to a mystical magic system.

(In an ideal world, it'd be cool if we could give these rooms, like, magic keywords to go to separate from the roomIds, but probably not gonna happen :P)

PR #309, opened 2020-10-02
Youtube embed #238

### LINKED ISSUE #238: Switch to YouTube Live embedding in theatre
Do we want to do this? (Just close the issue if not) - but it could be good for YouTube's more robust desampling support. 

https://webcam.io/support/howto-embed-youtube-live/ has some instructions. Note we'll want the "permanent URL" version.

kawa-kitsuragi:
Swapped the Twitch embed for the Youtube one: closes #238 .


MoyTW (approved):
Also needs to go into the popout HTML - see https://github.com/lazerwalker/azure-mud/blob/a0021b05a5e8d6ff84c10774dc3ecb85924f095e/stream.html

PR #316, opened 2020-10-02
Triage phantom room presence

MoyTW:
Doesn't fix us getting into the phantom room presence state but it'll
eventually boot them out on heartbeat.

Screenshot from 2020-10-02 14-35-47
Screenshot from 2020-10-02 14-36-07

### DIFF
--- a/server/heartbeat/index.ts
+++ b/server/heartbeat/index.ts
@@ -3,6 +3,8 @@ import { removeUserFromRoomPresence } from '../src/roomPresence'
 import { getFullUser } from '../src/user'
 import DB from '../src/redis'
 import { getHeartbeatData } from '../src/heartbeat'
+import { roomData } from '../src/rooms'
+import { globalPresenceMessage } from '../src/globalPresenceMessage'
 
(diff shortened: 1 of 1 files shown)


server/heartbeat/index.ts
```diff
@@ -57,12 +62,30 @@ const timerTrigger: AzureFunction = async function (
 
   await DB.setActiveUsers(activeUsers)
 
+  // This is a hack because sometimes people, for unknown reasons, get removed from activeUsers but *not* room presences
+  const roomIds = Object.values(roomData).map((r) => r.id)
```
lazerwalker, line 66: I think it's fine for now, since this won't meaningfully affect performance, but for your own knowledge, `Object.keys()` would give us what we want here (since the keys are the room IDs)

lazerwalker:
I think this is good!

If we run into performance issues (either Redis usage spikes or excessively long function execution time, neither of which I'm really worried about), low-hanging fruit would be to aggregate all of the `removeUserFromRoomPresence` calls for a single room into a single Redis call to update the array for that room instead of making 1 redis call per user.

PR #318, opened 2020-10-02
Colors are surprisingly hard

MoyTW:
...I don't think this is good enough, and I'm also baffled by why we're using `--green` for our links instead of the more traditional blue. But, I mean...it's better than it is now?

I went with making the highlight line really different than the background because that's how the other themes are and besides, we use it to outline buttons, where it's important for it to be different. I'm not sure why we use it to outline buttons though!

Screenshot from 2020-10-02 15-56-43
Screenshot from 2020-10-02 15-58-35
Screenshot from 2020-10-02 16-01-03
Screenshot from 2020-10-02 16-02-24
Screenshot from 2020-10-02 16-03-27
Screenshot from 2020-10-02 16-05-00
Screenshot from 2020-10-02 16-06-21
Screenshot from 2020-10-02 16-21-09

### DIFF
--- a/style/style.css
+++ b/style/style.css
@@ -8,16 +8,16 @@
   --gray-text: #b5b5b5;
   --alternate-background: #333333;
   --alternate-translucent: rgba(51,51,51,0.8);
-  --green: lightseagreen;
-  --highlight-line: #666666;
+  --green: #20b2aa;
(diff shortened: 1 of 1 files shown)


kawa-kitsuragi:
Going with the other solution instead, closing.

PR #324, opened 2020-10-03
Issue 322 Allow modal to scroll vertically

### LINKED ISSUE #322: Can't scroll down on map modal
When I'm in the map modal (this view):

Screen Shot 2020-10-02 at 6 00 05 PM

I can scroll side-to-side, but I can't scroll down to see the rest of the map. It looks like when I try to scroll up and down, the info _behind_ the map scrolls up and down instead of the map.

I'm using a scrolling gesture on a Mac laptop trackpad, using Chrome, browser window at about 1340x690.

It's not a blocker or anything since we also have the list.

MoyTW:
Also set its default max to big enough to see the whole map

closes #322

### DIFF
--- a/style/modal.css
+++ b/style/modal.css
@@ -28,8 +28,7 @@
 #modal.full-screen {
     /** This is a hard-coded value based on the current map size.
     A more automatic-updating system would be nice. */
-    max-width: 1000px;
-    max-height: unset;
+    max-width: 1300px;
 }
 
 td.time {


MoyTW:
Screenshot from 2020-10-02 18-38-26

kawa-kitsuragi (approved):
@lazerwalker should test for that weird vertical scroll behavior she saw before, but i wasn't seeing any problems. Merging for now.

PR #437, opened 2021-09-28
Switch to Firebase

MoyTW:
Proposed Firebase integration. Code is not clean.

**Open Questions:**
* Is the UI flow good?
* What's the best way to handle server secrets (specifically the firebase-admin.json file - see the README setup)
* Is importing the listener library for the verification page fine or are we trying to limit imports?

The pictures show the current email flow; up for UI consultation. The other flows open in a popup and ask you to log into the service.

I'm using `firebase-ui` to avoid having to build a UI, and it asks for the first & last name because Google. Could investigate changing that.

The email flow is wonky, it wants you to redirect back to the client (it's clearly designed with mobile-first in mind) but, since this is web, not client, you can't quiiiite do a deep link the same way, so I just redirect back to our app and have a "and if you're in the verification URL display a special screen" - we don't want two instances of the SPA app running around by default, that'd just be confusing. Open to discussion.

Currently when you go back to the "original" page you have to refresh. I'm thinking of adding a way to send a message to the other tab when you verify so you don't have to do that.

Screenshot from 2021-09-28 14-29-04

Screenshot from 2021-09-28 14-29-20

Screenshot from 2021-09-28 14-29-28

Screenshot from 2021-09-28 14-30-09

Screenshot from 2021-09-28 14-30-22

When you go back to the original tab you see this - maybe I should ask text saying "refresh" but what I'd really like to do is add broadcastchannel.

Screenshot from 2021-09-28 14-30-35

Screenshot from 2021-09-28 14-30-49


MoyTW:
For some reason, turning on email verification seems to have forced a password reset on _every_ login now?

lazerwalker:
Awesome, thanks so much!

Haven't looked at the code yet, but two immediate UI thoughts:

1. To make sure I understand the verification flow correctly: if I have the "we sent you a verification email" screen open, and then I click the verify link and see the "we verified you!" screen, does the first "we sent you an email" screen auto-update on verification, or does the user need to manually refresh the page? I have a few idea for clever tricks we can do, but I'm trying to get a better feel for how things work now.
2. On the "create name badge" screenshot, is that string of random letters and numbers the user-entered first name / last name, or some sort of Firebase-generated ID? If the latter, do we have access to the user's email address? That's what we populated Google users' usernames with last year (and Twitter handles for Twitter users), and while it's far from ideal, I think that's a more sensible default if possible.

MoyTW:
1) Between the PR authoring and now I added code to auto-update the first
screen when you revisit it
https://github.com/lazerwalker/azure-mud/pull/437/commits/e19493f28bf2233a5f128d699f59aec32b599d03
but I'm sure there's some way to get around that so I think adding a "and
if you have verified your email and you're still seeing this page refresh"
message would be good probably.
2) We do have access to the email address, we can do that.

lazerwalker:
Re: 1, that makes sense. I think that, with auto-updating and presumably adding a "refresh if it's not working" message, having people immediately close the verification window and treating the first window as the "proper" version of the app they have loaded seems like the best choice we have for now without a ton of effort.

src/components/MenuButtonView.tsx
```diff
@@ -17,12 +19,17 @@ export default function MenuButtonView (props: { username: string, spaceIsClosed
 
   const logOut = () => {
     const prompt = confirm('Are you sure you want to log out?')
+    // TODO: When a user disconnects via the /disconnect endpoint, it still shows that user as in the room. I suspect
```
lazerwalker, line 22: Have you tried this recently / with the latest changes from main? I pushed some changes on Wednesday 9/22 that I would have expected to fix this, so I'm curious if they're still persisting.

src/components/VerifyEmailView.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface Props {
+  userEmail: string;
+  dispatch: any;
+}
+
+const actionCodeSettings = {
+  // This URL should be changed to the frontend location for production. It must also be added to the authorized
+  // domains list in the Firebase Console.
+  url: 'http://localhost:1234',
```
lazerwalker, line 16: Just adding a comment to make sure this gets changed before merging

lazerwalker:
I looked through the code as it exists now, and it generally looks good! Not sure what you're thinking in terms of how much more clean-up/etc work is needed or not, but I'm broadly happy. Nice work :)

To explicitly answer your open questions:

1. I'm pretty happy with this now? If we add the "uh, refresh if it didn't auto-refresh" message and also get the profile screen auto-filling with email instead of a unique ID, I think this is totally reasonable.
2. I'm open to alternate ideas, but my instinct is that we .gitignore that JSON file, put the contents of the production JSON file into a GitHub Secret, and then the "deploy to production" GH action step has a line or two of bash scripting to write the contents of that secret/env variable to the correct file on disk before running the Azure Functions deploy. Thoughts?
3. Yeah, I'm fine with this. Page visibility APIs are slightly tenuous in that some content blockers aggressively block access to them for privacy reasons, but I'm okay saying that somebody doing that can just refresh the page (and there are probably better places to spend our time than implementing an alternate system for live-updating that page)

MoyTW:
Cleaned up. Fixed 1, 3.

As for 2, that sounds good - can you do that? I'm not familiar with github actions. Also as far as the actual, mechanical firebase setup, we should discuss that.

lazerwalker:
Merging this in. It's possible this will Just Work, but it's more likely I'll be manually tinkering in the `main` branch once it's in.

Issue #441, opened 2021-10-04
Video chat duplicates the feed when the user repeatedly opens settings


MoyTW:
When the user opens the settings, it will duplicate the feed on the screens of everybody else in the room.

Repro steps:
1) User A and User B are in the same room
2) User A opens the settings via either gear; nothing will happen yet
3) User A closes the settings
4) User A opens the settings a second time; this will duplicate the feed on User B's screen
5) User A closes the settings
6) Steps 4-5 may be repeated for multiple duplicate screens

Once you're in the duplicate state, if User A presses the Leave button it doesn't leave.

You can stop this issue by refreshing User A's browser.

MoyTW:
Oh, should've linked my PR. closing #443

PR #443, opened 2021-10-04
Issue 441: Fix duplicate chat context bug

MoyTW:
For issue #441 

Tested using the live server at chat.roguelike.club with a modified client, Chromium on Linux, logged in with different accounts on normal and incognito modes.


lazerwalker:
Thanks! 👍🏻

PR #451, opened 2021-10-04
Issue 442: Fix initial video visibility

### LINKED ISSUE #442: If a user refreshes the page then joins video chat, it is not visible to others immediately
User A and User B are in a room.

User B is not in video chat.

User A refreshes the page.
User A joins video chat.

At this point, User B cannot see User A's video chat.
Also voice transcription doesn't seem to work for User A.

If *either* user A or user B sends a message, the video chat immediately pops up on User B's screen.
If User A leaves and rejoins video chat, it works as intended.

This also seems to (sometimes) occur if User A is in a room, then leaves and returns to the room.

MoyTW:
So, this appears to fix the issue, but I'm not sure it's _how_ we'd like to fix the issue, given how inelegant it is.

Would appreciate commentary on it.

Closes #442

### DIFF
--- a/src/Actions.ts
+++ b/src/Actions.ts
@@ -42,6 +42,7 @@ export type Action =
   | MediaReceivedSpeakingDataAction
   | StopVideoChatAction
   | ErrorAction
+  | RefreshReactAction
   | SendMessageAction
   | SendCaptionAction
   | SetNameAction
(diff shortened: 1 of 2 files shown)


lazerwalker:
I suspect the fact that setRemoteParticipants doesn't trigger a repaint is itself an underlying smell, but 👍🏻 for "fixes this for now"!

lazerwalker:
I'm curious to test if this also sorts out my audio issues.

PR #454, opened 2021-10-05
Issue 452: Filter out no-publishing users from media view

### LINKED ISSUE #452: Joining and then leaving video chat still leaves the nametag
User A and User B are in a room.

User A joins video chat.
User A leaves video chat.

User A's nametag is still shown in the media view, whereas it should be expected that they are gone.

MoyTW:
Now no longer displays users who are in the Twilio room but are not connected to either audio or video tracks.

closes #452

### DIFF
--- a/src/components/MediaChatView.tsx
+++ b/src/components/MediaChatView.tsx
@@ -32,15 +32,34 @@ export default function MediaChatView (props: MediaProps) {
 
   let otherVideos
   console.log(callParticipants)
-  if(callParticipants) {
-    otherVideos = (Array.from(callParticipants.values())).map((p) => {
-      return (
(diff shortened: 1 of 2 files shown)


lazerwalker:
Thanks for this!

In an ideal world, I think a more idiomatically React version of this would probably push the logic down to the individual component — the div containing the NameView and media would be split out into its own React component, the `MediaChatView` would naively render all `callParticipants` as that component (even ones without audio or video tracks), and each individual instance of that component would be responsible for figuring out whether it should render `null` or actual content.

That said, a week out from the conference, I'm 100% happy with code that works, and willing to merge this in as-is if there's better stuff to be working on or if you're approaching burn-out.

PR #458, opened 2021-10-06
Issue 445 video move toggle

### LINKED ISSUE #445: Add settings toggle to auto-join videochat when entering a room
A toggle (defaulted to off) that sends your previously-selected audio and video tracks whenever you enter a new room. Presumably this lives in the same settings menu that theme selection is in, but maybe exposing it in the AV settings screen also makes sense?

MoyTW:
Adds a toggle that boots you out of the room every time you move rooms if you decide you want that.

One thing to note is that it actually calls `RoomView.leaveVideoChat` which is a little silly because it joins, publishes, and then immediately unpublishes. I am amenable to the clear refinement that hooking in the "don't publish in the first place" would be, but also don't know how much time we have.

Here's how it looks in settings:
Screenshot from 2021-10-06 16-47-58

Here's how it looks in audio settings:
Screenshot from 2021-10-06 16-47-42

And if you squish the audio settings down it's a mess - CSS is magic, idk.
Screenshot from 2021-10-06 16-48-12

closes #445


MoyTW:
As in "I don't think we should ship with video persisting between rooms" or "I don't think we should ship with the option to turn video off"?

lazerwalker:
So, the current behavior on `main` is "whenever you move to a new room, you have to explicitly join video. There are three possible viable options: leave that behavior as-is, replace that behavior with "if you're broadcasting video and move to another room, you automatically broadcast video in the new room", or have a user-configurable toggle. 

If a user does not have auto-join enabled (either because it's the only option, or because they were given the option), it's not acceptable that everyone in that room might get a flicker of their camera feed. 

Obviously the ideal case is one of us threads all this logic through to the twilioChatContext and letting users choose will work as intended. If that doesn't happen, I think whether the default behavior should be auto-join or not auto-join is a tough decision that's regrettably hard to say definitively without user testing.

PR #464, opened 2021-10-09
[WIP] Issue 444: revamped videochat UI

### LINKED ISSUE #444: New users in video chat jump your room around
If enough people join video chat (more than, like, 2? 3?) it bumps the room content around.

This is an unfun user experience.

MoyTW:
Open for UI consultation; frontend's not my forte. Also I'm not sure if this will cause big perf costs or not.

Right now what I've done is made the videochat a fixed-size element. It will always take [rows * video height] space up, and you can crank it up/down with the buttons. There's no ordering to the videos other than that your square is always first.

Screenshot from 2021-10-08 17-21-01
Screenshot from 2021-10-08 17-20-49
Screenshot from 2021-10-08 17-20-34
Screenshot from 2021-10-08 17-19-50
Screenshot from 2021-10-08 17-19-33

closes #444

### DIFF
--- a/src/components/LocalMediaView.tsx
+++ b/src/components/LocalMediaView.tsx
@@ -33,7 +33,7 @@ export default function LocalMediaView (props: Props) {
 
   return (
     <div className="my-video">
-      You:
+      You
       {localStreamView}
       {props.hideUI ? '' : (
         <div>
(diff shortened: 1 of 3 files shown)


lazerwalker:
Cool, thanks for this!

I've got a bunch of concerns about the implementation, but I think this is 100% fine for testing today, and then we can revisit.

PR #473, opened 2021-10-10
clickable map

kawa-kitsuragi:
- Map drawing now in Playscii, which is open source, cross platform, and by a past speaker!
- updated map, clickable regions, and presence list

plus some miscellaneous room changes to support - renaming a few rooms, making sure hashmark dungeon is as hidden as it should be


lazerwalker:
Thanks!

A few small things:

1. As mentioned in Discord, we need the room numbers to be 0-indexed instead of 1
2. Two changes here from the old map are (1) not using periods as 'floor' tiles in rooms, and (2) the placeholder numbers being 5 digits instead of 2 digits (00001 vs 01). Just wanted to confirm both of those are working as expected in the actual space. (
3. I'd probably put a minor vote towards reinstating the `.`s as floor tiles within rooms, but I'm not gonna dictate it if you think it looks better with spaces

PR #520, opened 2021-10-14
Space styling iteration + Riddle Modal

apepers:
The majority of this is styling of the space - room descriptions, the Map to match those rooms, expanding the generator text.

The only more 'functional' change is the addition of a Riddles field to the Room data, which if present will cause a button to be in the room details, which will open up the new Riddles Modal. It works in the same way as the modals for the colour changing doors, but with less logic because it just needs to pipe those riddle strings through into a modal popup and give them a bit of styling. 

There are still at this point placeholders in room descriptions for the riddle rooms - they're being iterated on now, your call on when you'd want to do a merge.


src/components/RiddleModal.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+export default function RiddleModalView (props: {riddles: string[]}) {
+  const riddleViews = props.riddles.map(function (val, index) { return <p className='riddle' key={index}>{val}</p> })
```
lazerwalker, line 4: Quick thing to note here! React expects keys to be globally unique, so just using the index as a number probably is going to lead to it yelling at us and maybe causing unexpected behavior.

An easy fix would be to make the key something like `{\`riddle-${index}\`}`, so it's still based on the index but won't clash with anything else

apepers: Thaaank you, that whole bit caused me quite a bit on confusion because I thought it was the attempt to iterate through the strings that was causing an error regarding it being missing a 'key' property. Let me make that change and push it.

PR #522, opened 2021-10-14
Font Reward

apepers:
This doesn't have all the riddle room text yet, but does have the wild chaos pit (which needs to be shortened or broken up into multiple rooms) - which is where I was able to hook up the font rewards. This has all the functional changes to get font rewards working and displaying, based off the same flow as the way we do user colours.


server/src/rooms/oldHub.ts
```diff
@@ -1,133 +1,124 @@
...
-  hasNoteWall: true,
-  hidden: true,
-  noMediaChat: true,
-  noteWallData: {
-    roomWallDescription: 'Do you want your possessions identified?',
-    noteWallButton: 'y/n/q',
-    addNoteLinkText: 'Tell a story',
-    addNotePrompt: 'Tell us of yet another sad/silly/surprising death you\'ve experienced in any roguelike.',
-    noteWallDescription: 'Tales of sad, silly, and surprising deaths in roguelikes of all kinds!'
+export const oldHubRoomData = {
```
apepers, line 3: The thought of duplicating all these room IDs in the index file made me sad so I changed it to actually let me export them all at once

PR #529, opened 2021-10-14
Based off PR #522 - not sure how to get GitHub to PR against it

MoyTW:
Based off https://github.com/lazerwalker/azure-mud/pull/522

https://github.com/lazerwalker/azure-mud/compare/ea96b96..feaa92a

The issue is that Alexei is working off another *repo* and I don't really know how to reconcile this, so this is MR'd against main, but you should review it against Kawa's last commit. There is always the concern of "but how will people know about this" oh well

Screenshot from 2021-10-14 15-45-17
Screenshot from 2021-10-14 15-45-25
Screenshot from 2021-10-14 15-45-30
Screenshot from 2021-10-14 15-45-35

### DIFF
--- a/src/Actions.ts
+++ b/src/Actions.ts
@@ -55,6 +55,7 @@ export type Action =
   | HideSideMenuAction
   | DeactivateAutoscrollAction
   | ActivateAutoscrollAction
+  | SetUseSimpleNamesAction
   | SetKeepCameraWhenMovingAction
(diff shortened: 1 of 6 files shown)


lazerwalker:
ohhh, hmm, I didn't realize there were newer commits from their past PR than are now in main. Going to move discussion to chat.

Issue #531, opened 2021-10-14
Public /go messages "spoil" hidden rooms


kawa-kitsuragi:
The current behavior is that everyone can see when a user leaves a room, and where they're going; you can also see where a user came in from.

This is great and maybe even ideal in public spaces, but does spoil the riddle puzzles, if you're standing in front of a riddle and see someone /go [solution]! 

Solution is to give an alternate message if someone `/go`-es to/from a room where `hidden = true`. 

![image](https://user-images.githubusercontent.com/62489292/137410187-4a4a4177-7df2-42cf-8324-dc6ee3a402f5.png)

apepers:
Closed with merged in change

PR #532, opened 2021-10-15
If a room is hidden, don't show links to it when a user enters/exits

apepers:
If someone leaves your room and goes to a hidden room, we shouldn't show a link to it, as that spoils the riddle games
Closing Issue #531

### DIFF
--- a/src/components/MessageView.tsx
+++ b/src/components/MessageView.tsx
@@ -29,6 +29,7 @@ import {
 import NameView from './NameView'
 import { DispatchContext, UserMapContext } from '../App'
 import { deleteMessage, fetchProfile, moveToRoom } from '../networking'
+import { roomData } from '../../server/src/rooms'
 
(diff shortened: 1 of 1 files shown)


src/components/MessageView.tsx
```diff
@@ -29,6 +29,7 @@ import {
 import NameView from './NameView'
 import { DispatchContext, UserMapContext } from '../App'
 import { deleteMessage, fetchProfile, moveToRoom } from '../networking'
+import { roomData } from '../../server/src/rooms'
```
apepers, line 32: I'm not sure if there's a better way to handle this part, or if there's a worry about referencing a server file from this more client side. It does seem to behave properly.

lazerwalker:
The failing lint check is you're not pulling in `roomData` from the context in `LeftView`, adding in a ` const roomData = useContext(RoomDataContext)` like the other components should work fine.

Fix that up, and swap out the server import for `state.roomData` in App.tsx, and this looks good!

Issue #553, opened 2021-10-15
Consider adding alternating background colors to each chat message


lazerwalker:
To increase readability.

Issue #552, opened 2021-10-15
Make enter/leave messages slightly more faded to reduce attention draw


lazerwalker:
Potentially just a blanket `opacity: 0.8`?

PR #566, opened 2021-10-16
Message readability

apepers:
Address issue #553 and #552 by making enter/leave messages dimmer, and giving every other message a different background colour


apepers:
Damn it 🤣 I saw your solution had also been to return null and though oh it'll just ignore my changes on that front, didn't notice the 'else' part.

PR #600, opened 2022-06-06
[WIP] Removing local dev dependency on Firebase

lazerwalker:
Just wanted to open a PR instead of this just sitting on a branch, not ready to merge yet. Slowly abstracting out our auth code so a local client can have an "eh, I don't really want to deal with Firebase" option.


lazerwalker:
This turned into a combination "let dev not use firebase" and "poke at how to get a local server working". I think it all turned out to be more work than I want right now, compared to just setting up a new dev instance myself with the current flow. Leaving the branch open but closing the PR.

Issue #593, opened 2022-04-26
Only show videochat 'join' button if user is a mod or a speaker


lazerwalker:
The logical way to do this is to check for the existence of either the 'mod' or 'speaker' badge, which means this is blocked on #591.

"Join" maybe the wrong word here. People should be automatically entered into the Twilio room when they enter a videochat room, but should not be allowed to enable broadcasting their own webcam/mic.

lazerwalker:
Logic exists to gate this. Logic to determine if someone is a speaker does not exist yet.

lazerwalker:
Worth noting we probably want to complicate this even more. If we have a game showcase, we probably want to enable audio-only streaming in the game showcase rooms. Letting people audiochat sounds fun, but bandwidth use is probably sketchy enough we don't want both a rainway connection AND a Twilio connection

lazerwalker:
This is blocked on https://github.com/Roguelike-Celebration/azure-mud/issues/658 and deciding if we actually have videochat. If we do have it, this is then also dependent on our "speaker" flag working

lazerwalker:
RIP old videochat

PR #602, opened 2022-06-15
Videochat design changes for 2021 [Closes #592][#593]

### LINKED ISSUE #592: Swap videochat flag in room data
Right now, it's assumed rooms have videochat unless `noMediaChat` is set to true. Flip that: rooms have no videochat by default unless explicitly enabled

lazerwalker:
Previously, the default behavior was that all rooms had videochat, unless a `noMediaChat` property was set to true. This flips that: by default, rooms do not have videochat, and they need `"mediaChat": true` set to enable it.

In 2022, the current design plan is to only enable videochat for mods and speakers in the speaker breakout rooms. This design will better facilitate that.

Additionally, this makes it so that, while all users can "join" videochat (in the sense that their client will receive others AV streams), only admins can actively broadcast their own audio/video. We will eventually expand this to include speakers, but we do not yet have an in-engine way to determine which users are speakers.

### DIFF
--- a/server/src/rooms/data/roomData.json
+++ b/server/src/rooms/data/roomData.json
@@ -17,7 +17,7 @@
     "displayName": "Endless Oatmeal",
     "shortName": "Oatmeal",
     "hidden": true,
-    "noMediaChat": false,
+    "mediaChat": true,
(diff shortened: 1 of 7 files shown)


lazerwalker:
It's worth noting we may need finer-grained controls around when videochat is allowed by non-admins -- @MoyTW mentioned the idea of a "use at your own peril" room people can go to for videochat, and if we eventually do some sort of integrated games showcase we may want to allow audio and/or video chat in there for spectators. But I don't think there's a reason to block merging this for now.

PR #604, opened 2022-07-01
[WIP] Add support for Rainway streaming

lazerwalker:
This adds a bunch of things:
- Updates React from 16 to 17 (necessary for the Rainway library)
- Replaces react-contextmenu with our own fork that does nothing different (react-contextmenu is abandonware and is hardcoded to require React 16. We will probably want to rip this out and replace it)
- Give the admin page the ability to create new rooms
- Add in a Rainway DOM object when a hardcoded roomId is hit, that's basically the demo sample app (i.e. you have to manually paste in an API key and peerID). This is the bit we'll remove and refactor if we keep this.

This requires a companion app running on a PC, which is currently just their Node sample. A prod deployment willw
 require customizing it to add process IDs to an allowlist, and investigating whether we'd get better performance from C#.

The scope of this is letting us test how well Rainway works on Sunday. If successful, this will probably be the basis of a production-ready Rainway code installation, but this definitely isn't there yet (what's missing: those hardcoded bits, being able to specify associate specific streams with specific rooms, our logic for handing off control between players, an allowlist for game process IDs and a process for updating those)


lazerwalker:
Merging this in as soon as the tests go green. If you `/move gamerCave`, you'll have access to a copy of the Rainway React sample lifted basically wholesale (minus p2p chat). This is far off from our desired implementation, but will let us test it in our Sunday meeting today.

PR #605, opened 2022-07-04
Store Rainway API key in ENV

lazerwalker:
This removes the "enter API key" field when testing out Rainway.

There's a `rainway-2` branch that has deeper integration for sending Rainway peer IDs over the wire, but because that involves messing with the rest of the app's infrastructure (a HTTP endpoint, a new SignalR message, a new Action, etc) I'm holding off on merging that in until we know we want this, since right now this is a clear revert.


lazerwalker:
This requires a new value to be set in GH repository secrets; this repo currently has the correct key set. Going to merge once it's green.

PR #608, opened 2022-07-24
Enforce required node version

rmkubik:
Added a few configuration options to enforce Node.js 16 as required node version in this project. 

- Added Node and NPM versions to `package.json` & `package-lock.json`
- Set `.nvmrc` to use node 16. 
- Set `.npmrc` to use strict-engine. I haven't used this setting before, but it causes `npm` commands to fail if the current environment is not running the node versions specified in `package.json`. I could see this having unintended side effects as I do not know the full ecosystem of this project though! A bespoke "preinstall" script with a console warning vs. hard failing might be preferred!

When onboarding, I was on node 14 and this caused the app to crash without indicating I was on the wrong version. Hopefully, enforcing the appropriate version in this way will improve the onboarding experience for future developers.


lazerwalker:
Awesome, thanks!

Issue #155, opened 2020-08-25
Visually differentiate between messages in current room vs messages received in previous rooms


lazerwalker:
Preview event feedback: "Transitioning rooms I keep expecting my chat in a new room to be blank because I’ve moved. The way it stays the same makes me think I can still hear people in the previous room. Unsure if this is standard MUD convention since I didn’t actually play MUDs -  I think my Discord/Slack/etc expectations are why I think new room = new text"

lazerwalker:
Yeah, I agree! This is super tricky! Traditional MUDs get around this by having the room description basically clear out previous messages in the inline chat scroll, but that doesn't really work with our design. 

A few thoughts of what might work:
* Hide previous chat, with a “show previous messages” button that loads up the full chat scroll
* We still show the full chat scroll when you move, but there’s a stronger visual indicator (e.g. a horizontal line or something) indicating “hey, new room!”
* Both of these. Messages before the current room are hidden behind a button, but there’s still horizontal lines to indicate previous room shifts

lazerwalker:
I'm leaning towards option (c). Messages before the current room are hidden behind a button, and if you click it then we load them all in, with `<hr>`s delineating movement.

lazerwalker:
Not super high-priority, but a quick client-side-only thing.

lazerwalker:
Happy to close this for now, even if we end up tweaking behavior

PR #609, opened 2022-07-27
Show older messages from previous rooms behind a toggelable button

rmkubik:
This is a start at an implementation for #155.

- This PR adds an `hr` into the chat log before every `MovedRoom` message
- This PR adds a button-link to the top of the chat panel to toggle the showing/hiding of messages from rooms older than the current one

Screen Shot 2022-07-26 at 10 11 54 PM

Screen Shot 2022-07-26 at 10 11 49 PM

Design-wise, I'm not so sure about the positioning of the "Show/Hide older messages" link. I think the "Hide older messages" version is particularly confusing as implemented here. Changing the wording might solve this issue. Something like "Hide messages from previous rooms"? But then the link starts getting pretty long!

The other idea I had design-wise was to insert the link into the chat window by the `hr` for the most recently entered room. This feels like a pretty drastic vibe change though, so didn't really want to pursue that on my own. Additionally, if a user has scrolled pretty far up in past messages or if they have received a lot of current messages they'd lose the reference to the "show older messages" button. But maybe this is desirable?

Implementation-wise, I have added a few comments. I'm not sure if my approaches are inline with the project's architecture or not!

### DIFF
--- a/src/components/ChatView.tsx
+++ b/src/components/ChatView.tsx
@@ -89,15 +89,20 @@ export default function ChatView (props: Props) {
           }
         }
 
+        const shouldShowInterstitial = m.type === MessageType.MovedRoom
         const id = `message-${idx}`
+
         return (
-          <MessageView
-            message={m}
-            key={id}
-            id={id}
-            hideTimestamp={hideTimestamp}
-            msgIndex={idx}
-          />
+          <>
+            {shouldShowInterstitial ? <hr key={id + '-interstitial'}/> : null}
+            <MessageView
+              message={m}
+              key={id}
+              id={id}
+              hideTimestamp={hideTimestamp}
+              msgIndex={idx}
+            />
+          </>
         )
       })}
     </div>
--- a/src/components/ChatView.tsx
+++ b/src/components/ChatView.tsx
@@ -1,4 +1,5 @@
 import React, { useContext } from 'react'
+import { findLastIndex } from 'lodash'
 
 import MessageView from './MessageView'
 import { Message, MessageType, ConnectedMessage, DisconnectedMessage, EnteredMessage, LeftMessage } from '../message'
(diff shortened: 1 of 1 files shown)


src/components/ChatView.tsx
```diff
@@ -48,6 +48,8 @@ export default function ChatView (props: Props) {
     }
   })
 
+  const [shouldShowOldMessages, setShouldShowOldMessages] = React.useState(false)
```
rmkubik, line 51: I see a lot of context/reducer usage in this app to store app state. This felt like local state that the rest of the app outside chat didn't need knowledge. Not sure if that's inline with the state management strategy here though!

src/components/ChatView.tsx
```diff
@@ -67,39 +70,60 @@ export default function ChatView (props: Props) {
...
-          />
-        )
-      })}
-    </div>
+          const shouldShowInterstitial = m.type === MessageType.MovedRoom
+          const id = `message-${idx}`
+
+          return (
+            <>
+              {shouldShowInterstitial ? <hr key={id + '-interstitial'}/> : null}
```
rmkubik, line 114: I opted to add this interstitial logic instead of changing the `MovedRoomView` here. The `MessageView` component doesn't really allow a message to interject an `hr` before the message. 

---

The other approach I considered was creating a dummy `MessageType` called something like `MessageType.Decorator` that took in some React content. Then we could interject decorator messages where we liked in the `messages` list for `hr`'s or other elements in the future. 

This also didn't necessarily feel right since these things weren't really messages, and it felt weird to extend the `MessageType` type to accomplish this approach.

---

Open to input on the "correct" approach for this code base.

src/components/ChatView.tsx
```diff
@@ -48,6 +49,8 @@ export default function ChatView (props: Props) {
     }
   })
 
+  const [shouldShowOlderMessages, setShouldShowOlderMessages] = React.useState(false)
```
rmkubik, line 52: I see a lot of context/reducer usage in this app to store app state. This felt like local state that the rest of the app outside chat didn't need knowledge. Not sure if that's inline with the state management strategy here though!

lazerwalker (approved):
This looks great, thanks! 

I agree figuring out what to do with the "hide" button is a bit tricky. I think if we wanted to play with showing the button inline next to the hr, we'd probably want to default to hiding all previous messages whenever you move rooms (and I guess maybe make that toggleable in the settings?). Which all seems doable, and I think that shift would also minimize the "urgh, this room is too busy and now I can't find the show/hide button" problem. 

I'm also okay with saying that the simple thing that already works is fine, and if either we start to hate it in internal testing or if people are confused during the preview event, that's a great chance to revisit.

rmkubik:
I'm happy to ship as is then. Perhaps living with it will make clearer what a better solution is for the button itself. 

I'm curious how it will feel in situations with more traffic.

PR #610, opened 2022-07-31
[WIP] Virtualize Messages with `react-window`

mysterycommand:
After looking at `react-virtualized` it turns out that author rewrote that with a smaller/more focused feature set as [`react-window`](https://react-window.now.sh/). It seems to handle everything we'd want it to, so this is a start at implementing that.

So far it's been pretty smooth, but I'm kinda stuck scratching my head at an issue where the `FixedSizeList` appears to render extra space at the end of the list. I'm still not sure where that's coming from, tbh … also, I've noticed issues with longer messages and I think I need to switch to `VariableSizedList` and keep some kind of array of list item sizes as they render (though maybe these would need to be invalidated if a user resizes their screen?)

Anyway, I wanted to post some progress before the meeting, so here it is


MoyTW:
2 years old, likely out of date, we need to do better about groom our PRs.

PR #622, opened 2022-09-03
Issue 327 calculated

MoyTW:
Fixes the browser drift issues on the map by more precisely calculating the divs.

I'd do pictures but it's 1:20am and I need to actually get up this weekend, what am I doing.


lazerwalker:
Love it!

PR #638, opened 2022-09-10
Adds Generated Drinks

RourkeBywater:
Allows the player to order a newly generated drink at the Bar. orderNewDrink is essentially a mashup of displayMessage and pickUpItem, and it uses four new tracery grammars: drinkSkeletons, drinkVessels, drinkContents, and drinkNames. I am not confident I did everything right here.


server/src/endpoints/orderNewDrink.ts
```diff
@@ -0,0 +1,57 @@
...
+  thisDrinkSkeleton = drinkSkeletons.generate()
+
+  thisDrinkSkeleton = thisDrinkSkeleton.replace('-drinkVessel-', thisDrinkVessel)
+  thisDrinkSkeleton = thisDrinkSkeleton.replace('-drinkContent-', thisDrinkContent)
+  thisDrinkSkeleton = thisDrinkSkeleton.replace('-drinkName-', thisDrinkName)
+  thisDrinkSkeleton = thisDrinkSkeleton.replace('-userName-', user.realName)
+
+  const item = thisDrinkName
+
+  const actionString = (item ? `picks up ${item}.` : `drops ${oldItem}.`)
```
lazerwalker, line 30: If someone doesn't have an item, this will generate a "so-and-so picks up [drink]" message, but if they do have an item, it will just say "so-and-so drops [old item]", with no reference to the new item they have. Just want to confirm that's the behavior you want instead of the "drops" message to also reference the new item.

server/src/endpoints/orderNewDrink.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const newProfile = await updateUserProfile(user.id, { item })
+
+  const messages: Message[] = [
+    {
+      userId: user.roomId,
```
lazerwalker, line 36: For `actionString` and `thisDrinkSkeleton`, which ones are supposed to be private to just the player getting a drink and which ones are supposed to show to everyone in the room?

For "so-and-so is drinking a potion" commands, you probably want to use `emote` instead of `privateCommand` (and target the roomId). For a private "you are doing a thing" message, `privateCommand` does work, but you want the `userId` to be e.g. `user.id` instead of `user.roomId`. Sending it to `user.roomId` will send it to every user in the room instead of just that one person.

lazerwalker (commented):
Two minor things, otherwise looks good!

If I understand correctly, once merged in, this should theoretically make it so that ordering a new drink works 100% as expected, with zero "order someone else's drink" functionality (pending me helping with database stuff). Does that sound correct?

server/src/endpoints/orderNewDrink.ts
RourkeBywater: They're both supposed to go to everyone in the room. However, probably 'you are doing this thing' can just be private, I'll change it accordingly. Should I change thisDrinkSkeleton? I do want everyone to see that one.

RourkeBywater: Would chatMessage be a more appropriate target?

server/src/endpoints/orderNewDrink.ts
RourkeBywater: Oooh ok I will fix that right now!

RourkeBywater:
> Two minor things, otherwise looks good!
> 
> If I understand correctly, once merged in, this should theoretically make it so that ordering a new drink works 100% as expected, with zero "order someone else's drink" functionality (pending me helping with database stuff). Does that sound correct?

That's correct! I would like to have order someone else's drink functionality, but I want to make sure this works first and I don't think there is time for the other stuff before tomorrow.

server/src/endpoints/orderNewDrink.ts
RourkeBywater: I'm actually not sure how to make that happen? I'm using the code from pickUpItem for this.

server/src/endpoints/orderNewDrink.ts
lazerwalker: "chatMessage" is what we use for a normal text chat message you've sent, that probably isn't appropriate.

What do you want this to look like?

RourkeBywater: I want it to look like a thing happening in the room? I don't know what would be right for that.

lazerwalker: "A thing happening in the room" = an italicized message that isn't a specific player doing an action?

I think in that case `privateCommand` sent to the entire room is the only tool we have, even though it's named incorrectly / there's no guarantee it will continue to look that way in the future since this isn't what it's intended to be used for.

RourkeBywater: I've been looking through the other endpoints and haven't found one that seems to do something similar? The closest I've found is something like displayMessage for when talking to the bartender, but that's a private command.

RourkeBywater: Yea, that's what I'm looking for it to be. Changed to to privateCommand for now.

server/src/endpoints/orderNewDrink.ts
RourkeBywater: Still trying to work this out, I don't think I understand the syntax. I would like it to do both, as with the normal pickUpItem.

lazerwalker: Ohhh, I'm being silly and misread this. The behavior here is "if the user is not trying to pick up an item (read: they clicked the "drop item" button), tell them they dropped their old item. Otherwise, quietly drop the old item and display they grabbed the new one"

RourkeBywater: Ooooh, ok, that makes sense.

lazerwalker: In case that syntax is confusing, it's called a "ternary". A statement that looks like `x ? y : z` is roughly equivalent to `if(x) { y } else { z }`.

(In JS, if you try to take something that isn't a boolean, like the `item` var here, and ask if it's `true`, the rules are generally "does this thing exist / is it defined? if so, it's considered `true`", with some weird unexpected exceptions like the number 0 is `false` and the string `""` is false)

Issue #664, opened 2022-09-26
Profile data maybe not applying


lazerwalker:
From the note wall: "I edited my profile and it's
overwriting my pronouns with the
contents of the IRL name field,
and then after submitting it
doesn't load the current field
values if I try to edit again
without reloading the page"

rmkubik:
It looks like this is happened because the [`updateProfile` networking.ts](https://github.com/Roguelike-Celebration/azure-mud/blob/main/src/networking.ts#L123) call is expecting the new `.user` to be attached to the result after it calls the Azure Function `updateProfile`.

However, even though the result comes back valid, there is no other data attached to the result.

I am not familiar with the server side of this app, but it seems like the [`updateProfile` API ](https://github.com/Roguelike-Celebration/azure-mud/blob/main/server/updateProfile/index.ts)does not seem to have any sort of return statement where the user would be appended.

I think I need a little more back end context to proceed here, or perhaps someone with a working backend set up should grab it. :-)

rmkubik:
Okay, so the original ticket does seem to be solved!

We seem to be correctly changing pronouns now after #680 is merged.

rmkubik:
The issue above however, with .user being undefined is still causing a different issue.

When you edit your profile and submit it and the request responds successfully no user is returned on the result object. We then set this undefined object to the user's state in the client side. This means, when you re-open the Edit Profile form all your data fields will be empty. Your profile will still be updated as expected if you view it. If you then submit the newly empty Edit Profile form, this will clear out all your profile data.

rmkubik:
Looks like #540 already addresses this remaining issue so I will go ahead and mark this issue as resolved!

PR #680, opened 2022-10-03
Don't set data.pronouns to data.realName

rmkubik:
I think this is part of the issue with #664. I'm not sure if it will entirely solve this issue though. 

This fix does not resolve the issue for me locally. However, I think this function I'm fixing is being used on the backend. I am using the production backend so not sure if the fix works yet.


lazerwalker:
Woof, thanks, I'm guessing this has been floating around since 2020!

And yep, if you're not running your own Azure backend instance, it's not expected you'd be able to test this before merging in. Seems like a low-risk merge.

PR #681, opened 2022-10-03
Adjust css on input/send and profile close

nextlevelbanana:
Fixed the weird character wrapping on the profile view SEND button, and did some other minor style tweaks while I was here: centered the X in the close button, centered the input fields and their SEND buttons, added a little bit of space in between field and button.

Looks like this now (ignore my lack of proper mod badge, apparently I can't be bothered to copy and paste tonight)
image

### DIFF
--- a/style/input.css
+++ b/style/input.css
@@ -9,6 +9,10 @@
   width: 100%;
 }
 
+#input .mentions {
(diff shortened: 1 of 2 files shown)


lazerwalker:
Love it!

PR #685, opened 2022-10-06
Update year to 2022 where it said 2021

nextlevelbanana:
Time is an illusion. Just updating some strings.


kawa-kitsuragi (approved):
Thanks!

PR #688, opened 2022-10-09
Messages Context

mysterycommand:
This PR turns `state.messages` into an object like `{ entities: Record<string, Message>, ids: string[] }` so that we can pass an array of ids into a messages list, which'll just pass a single id to a message item view, and then the message item view can "look up" the actual message object from a(n also new) `MessagesContext`. For right now it just updates the main reducer and state object and converts back to the old style of "array of message objects" so that the changes are pretty minimal (apart from formatting). I'll add some comments to the PR to add more detail around the interesting bits.


src/App.tsx
```diff
@@ -362,74 +363,76 @@ const App = () => {
...
+                            presenceData={state.presenceData}
+                            inMediaChat={state.inMediaChat}
+                            keepCameraWhenMoving={state.keepCameraWhenMoving}
+                            textOnlyMode={state.textOnlyMode}
+                            mediaChatView={videoChatView}
+                            hasDismissedAModal={state.hasDismissedAModal}
+                          />
+                        ) : null}
+                        <ChatView
+                          messages={Object.values(state.messages.entities)}
```
mysterycommand, line 415: This converts the new `messages.entities` object back into the expected array of message objects for the chat view.

src/reducer.ts
```diff
@@ -533,7 +660,15 @@ export default produce((draft: State, action: Action) => {
...
-    draft.messages = action.messages
+    draft.messages = action.messages.reduce(
+      (acc, message) => {
+        acc.entities[message.id] = message
+        acc.ids.push(message.id)
+
+        return acc
+      },
+      { entities: {}, ids: [] }
+    )
```
mysterycommand, line 671: This converts from the array of message objects stored in local storage into the "entity state" style, using the current state of messages as the starting point.

src/reducer.ts
```diff
@@ -620,30 +760,45 @@ export default produce((draft: State, action: Action) => {
 // WARNING: These three functions modify the message state without awaiting on the result.
 // If you're seeing weird race conditions with the message store, that's probably the issue.
 
-function deleteMessage (state: State, messageId: String) {
-  const target = state.messages.find(m => isDeletableMessage(m) && m.id === messageId)
+function deleteMessage (state: State, messageId: string) {
+  const target = state.messages.entities[messageId]
 
-  // Calling isDeletable again here so TypeScript can properly cast; if there's a nicer way to do this, please inform!
   if (isDeletableMessage(target)) {
```
mysterycommand, line 766: Only needs to do the `isDeletableMessage` check once which is kinda nice.

src/reducer.ts
```diff
@@ -533,7 +660,15 @@ export default produce((draft: State, action: Action) => {
...
-    draft.messages = action.messages
+    draft.messages = action.messages.reduce(
+      (acc, message) => {
+        acc.entities[message.id] = message
+        acc.ids.push(message.id)
+
+        return acc
+      },
+      current(draft.messages)
+    )
```
mysterycommand, line 671: This converts from the array of message objects stored in local storage into the "entity state" style, using the current state of messages as the starting point.

src/reducer.ts
```diff
@@ -533,7 +660,10 @@ export default produce((draft: State, action: Action) => {
   }
 
   if (action.type === ActionType.LoadMessageArchive) {
-    draft.messages = action.messages
+    action.messages.forEach((message) => {
+      draft.messages.entities[message.id] = message
+      draft.messages.ids.push(message.id)
+    })
```
mysterycommand, line 666: This converts the array of messages into the "entity state" style

src/reducer.ts
```diff
@@ -660,10 +660,17 @@ export default produce((draft: State, action: Action) => {
...
+    const nextEntities = {
+      ...current(draft).messages.entities,
+      ...action.messages.reduce((entities, message) => {
+        entities[message.id] = message
+        return entities
+      }, {})
+    }
+
+    draft.messages.entities = nextEntities
+    draft.messages.ids = Object.keys(nextEntities)
```
mysterycommand, line 672: This converts from an array of message objects to the entity state style, while making sure that we don't inadvertently duplicate any ids … messages coming in through the action should overwrite any messages that already exist in state (if all their values are the same this will wind up being a no-op on the Immer side)

src/reducer.ts
```diff
@@ -533,7 +660,17 @@ export default produce((draft: State, action: Action) => {
...
+    const nextEntities = {
+      ...current(draft).messages.entities,
+      ...action.messages.reduce((acc, message) => {
+        acc[message.id] = message
+        return acc
+      }, {})
+    }
+
+    draft.messages.entities = nextEntities
+    draft.messages.ids = Object.keys(nextEntities)
```
mysterycommand, line 672: This converts from an array of message objects to the entity state style, while making sure that we don't inadvertently duplicate any ids … messages coming in through the action should overwrite any messages that already exist in state (if all their values are the same this will wind up being a no-op on the Immer side)

lazerwalker:
This PR looks great in its current form! (modulo the reducer git conflicts that look trivial). As per your suggestion on GH, I'm going to avoid merging this in until I've had time to properly look over #2 and #3 on your fork so we can merge those into this as a single PR.

mysterycommand:
> This PR looks great in its current form! (modulo the reducer git conflicts that look trivial). As per your suggestion on GH, I'm going to avoid merging this in until I've had time to properly look over #2 and #3 on your fork so we can merge those into this as a single PR.

Fixed those merge conflicts 🤞 tests all pass, updating https://github.com/mysterycommand/azure-mud/pull/2 and https://github.com/mysterycommand/azure-mud/pull/3 now

MoyTW (approved):
Nothing obviously terrible jumped out to me and I want folks to bang on it

PR #695, opened 2022-10-10
Add past speaker badge and mod menu item

nextlevelbanana:
Now... we will have to go find all these past speakers in order to give them their badge. Or else they'll have to whisper us and ask for it. But hey, we'll have the option!

### DIFF
--- a/src/components/NameView.tsx
+++ b/src/components/NameView.tsx
@@ -6,8 +6,7 @@ import {
   BanToggleAction,
   HideModalAction,
   ModToggleAction,
-  ShowModalAction,
-  UnlockBadgeAction
+  ShowModalAction
 } from '../Actions'
 import { User } from '../../server/src/user'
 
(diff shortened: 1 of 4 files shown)


src/components/NameView.tsx
```diff
@@ -85,6 +86,28 @@ export default function NameView (props: Props) {
     }
   }
 
+  const handlePastSpeaker = (e, data) => {
+    dispatch(UnlockBadgeAction(data.badge))
```
lazerwalker, line 90: Have you tested this in a local web client? IIRC `UnlockBadgeAction` is responsible for showing the local user the "hey, you unlocked a new badge!" modal and showing that badge in the selection UI — I don't think this will trigger a network request to server-side unlock a badge for a different user.

lazerwalker:
I'm not sure this works as intended (but I'd love to be wrong!)

My gut instinct is we need to do one of two things:

1. Update the `setSpeakerStatus` endpoint to take in a year. If the year isn't 2022, that should probably assign the correct user the correct badge on the server side
2. Create a new endpoint that takes in a user ID (that isn't the logged-in user, who must be a mod) and assigns them a past speaker badge

#1 would probably be a fair bit simpler

src/components/NameView.tsx
nextlevelbanana: Ah dang, I think you're right. Which would explain why there's no user ID involved

nextlevelbanana:
Ok, I'll take another stab if I have time :)

lazerwalker:
This seems reasonable! Merging so we can test on prod 😂

PR #696, opened 2022-10-11
Send /mod messages to discord

nextlevelbanana:
I tested this on my own server with a hardcoded URL, so double-check I didn't typo the environment variable name. 

It did send every message _twice_, which isn't ideal. For some reason the reducer is running twice (and I'm not familiar enough with reducers to sort out why). Maybe it's because I myself am a mod? In any case I think double-messages are better than none, and if it turns out to be too spammy, we can always just disable the webhook.

### DIFF
--- a/src/reducer.ts
+++ b/src/reducer.ts
@@ -40,6 +40,8 @@ import firebase from 'firebase/app'
 import Config from './config'
 import { Badge } from '../server/src/badges'
 import produce from 'immer'
+import { sendToDiscord } from './sendToDiscord'
+
 export interface State {
   firebaseApp: firebase.app.App;
   authenticated: boolean;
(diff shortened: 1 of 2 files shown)


lazerwalker:
@nextlevelbanana I realized right after merging this in: the reason it's triggering multiple messages is the client `ModMessage` reducer action fires any time a client is sent a "mod" message from the server, which happens both for the person who sends a /mod message and for each logged-in mod.

We probably want this to happen exactly once on the server, probably somewhere in this block: https://github.com/Roguelike-Celebration/azure-mud/blob/fb856b33d97dd619adb87f0479434d13e6866666/server/src/endpoints/sendChatMessage.ts#L70

nextlevelbanana:
Aha! That could get real noisy real fast 😬 ok, I'll pick this up again tomorrow

PR #700, opened 2022-10-14
Address SignalR disconnects

MoyTW:


MoyTW:
Test this by turning off your internet or closing your laptop or just leaving the tab open for, like, an hour, which seems to semi-regularly disconnect me!?

Possible points of discussion:
+ Do we care about people leaving the tabs open forever, which will both perpetually reconnect and eat up cycles on their computer if their internet is bad? It's not many cycles, mind, but...
  - Oh actually I just remembered that we have a timeout - but I don't remember how it works, and also, I'm pretty sure it doesn't provide feedback to the client, which means that to the player it's invisible? uh, yeah, deserves discussion
+ By the time it reaches the SignalR disconnected callback you've probably already been disconnected for a good twenty, thirty seconds, and you'll drop messages in between since we have no server playback. This is...fine, right?

lazerwalker:
Awesome, this all seems reasonable!

Assuming you mean our "ping" system re: timeouts, I'm trying to remember how that works. IIRC it forcibly closes connections if the client is in a weird state where it's still connected to SignalR but isn't processing incoming messages. I'm not sure how that will or won't intersect with this (and seems separate from the question of idle clients continually trying to reconnect via this new modal)

I suspect an item to investigate for next year is shifting to Azure PubSub Service (a new similar service that uses vanilla websockets instead of SignalR, introduced in the last year), since we don't actually care about the complexity of SignalR -- we're only using SignalR because we want persistent connections to a serverless backend, which was previously only doable on Azure with SignalR Service. We're obviously not going to throw out our entire transport layer one week before the event, but calling that out here since it does seem like part of this jank is SignalR being so complex in ways that don't benefit us.

PR #703, opened 2022-10-15
Update schedule for 2022

lazerwalker:
- Adds a new schedule. 
- Adds some logic to automatically show the correct day of schedule (in the past, we pushed a commit after day 1 commenting it out and uncommenting out a previously-written day 2 schedule. Now, we check if it's at least 1 hour after the last event on day 1, and if so swap to day 2)
- Make a single place where we can fake out the current date to locally test both the schedule view and the "now happening" view
- The schedule now says whether it's day 1 or 2

### DIFF
--- a/src/components/HappeningNowView.tsx
+++ b/src/components/HappeningNowView.tsx
@@ -4,7 +4,7 @@ import { HideModalAction } from '../Actions'
 import { DispatchContext } from '../App'
 import { moveToRoom } from '../networking'
 import { Room } from '../room'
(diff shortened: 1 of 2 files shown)


lazerwalker:
lol I screwed up git bad.

lazerwalker:
Ugh, this accidentally added an empty RateTalkView.tsx file. Don't think there's any harm in it, will wait to delete it until it's clear whether we'll actually get that feature in or not.

PR #705, opened 2022-10-15
move sendToDiscord to the server

nextlevelbanana:
take two. Moved `sendToDiscord` to the server and called it from `sendChatMessage.ts`.

Things I'm not confident about:
- do we use "request" to post to the webhook? (what version of node is the server running? / could I have still just used fetch?) 
- do we need to add the discord url secret to deploy.yml? Currently I do not believe it is there.

### DIFF
--- a/server/src/endpoints/sendChatMessage.ts
+++ b/server/src/endpoints/sendChatMessage.ts
@@ -8,6 +8,7 @@ import { getUserIdForUsername, User } from '../user'
 import { MESSAGE_MAX_LENGTH } from '../config'
 import { dance } from '../dance'
 import { interact } from '../interact'
+import { sendToDiscord } from '../sendToDiscord'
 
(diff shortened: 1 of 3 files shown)


lazerwalker:
re: making HTTP requests, I have no idea. Let's see if this works 😂. If it doesn't work, I don't know what our server Node version is, but it looks we already have `node-fetch` as a sub-sub-dependency in our package.json, would probably be reasonable to pull that in.

Re: secret keys, I've just added the `DISCORD_URL` key (and the value from Discord chat) to the Azure app. If you need to tweak it for any reason, the Azure Function App has a list of secret values in the "Configuration" subsection of the Function App portal page. We would need to add it to deploy.yml if we wanted it to actually work from the client, oops (it would have worked for you locally assuming you dropped it into your local dev .env file)

PR #716, opened 2022-10-17
WIP: Migrate to ESM

lazerwalker:
This:
- Upgrades our TS version for the latest goodness
- Sets `type: module` to tell Node that all .js files are ESM
- Tells TypeScript to emit ESM instead of commonJS

Not currently mergable because Azure Functions doesn't seem to support `type: module`. A few options:
1. Write a manual build script to rename all `.js` files to `.mjs`, and remove `type: module`
2. Use node's default HTTP module instead of node-fetch
3. Give up on our Discord mod bot integration
4. Hope I get a better answer internally

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -22,7 +22,7 @@
         "@azure/functions": "^1.0.2-beta2",
         "@types/redis": "^2.8.24",
         "recursive-readdir": "^2.2.2",
-        "typescript": "^3.3.3"
+        "typescript": "^4.7"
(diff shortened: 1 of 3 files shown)


lazerwalker:
No action taken for 2022, Travis downgraded node-fetch.

For future-us, TS 4.7 includes support for `mts` files that compile down to `mjs`. Moving all of our files to mts/mjs should allow this to Just Work on Azure Functions.

PR #729, opened 2022-10-21
possible fix for room badge unlock

nextlevelbanana:
When you enter a room with a badge, it's supposed to show you the modal saying "hey, you got a badge!" but it hasn't been. I'm guessing this is why? I don't know, but I definitely can't break it any worse 🤙


MoyTW (approved):
trust in the power of revert

PR #744, opened 2022-10-22
Allow rendered @-mentions in emotes and whispers

lazerwalker:
Previously, if you @-mention someone in a message that isn't a normal chat message, it gets rendered as the underlying @[[username]] syntax instead of actually rendering a NameView.

This ports the name parsing/rendering logic to EmoteView and WhisperView. It's possible this should be added to another message type, but as a gut check this seemed fine.

Possible this will cause additional rendering load, but it shouldn't.

Screen Shot 2022-10-22 at 13 04 42


lazerwalker:
(That Sonat commit will hit main before we merge this)

PR #796, opened 2023-09-07
UI Updates: preview schedule, current year, no twitter in profile

nextlevelbanana:
- changed all instances of 2022 to 2023
- changed "twitter handle" to "social link" on profile (note: I changed the character limit from 20 to 200, to match our other URLs. Will that, uh, break a db table somewhere or am I good?) and made minor tweaks to the profile view UI
- added preview event to schedule view, updated schedule view with 2023 event dates

I don't have much experience working across forks, please double check I didn


MoyTW (approved):
lgtm

PR #801, opened 2023-09-10
Fix apos in Twitter note

lazerwalker:
Screenshot 2023-09-09 at 6 12 00 PM

This is just a small typo, fixing a few instances of `&pos;` to `&apos;`, the actual escape code.


MoyTW (approved):
Nice, thanks.

PR #811, opened 2023-10-10
Content Update

nextlevelbanana:
- Add starters for several generators (not hooked up to rooms yet)
- Change empty store to Spirit Crawloween
- Add room badges for Spirit, Hot Dog Stand
- Add unlockable Hot Dog Stand color theme

I haven't done anything with the roomData yet. I pulled down what's live in prod and compared it to what's in main and the diff was weird, so I decided I'd do that next.


MoyTW (approved):
didn't actually read this but it'll probably be fine!

PR #812, opened 2023-10-12
Sync room data

nextlevelbanana:
Saving room data from production
Apparently &apos; isn't the right way to escape apostrophes inside the generators, lol. Fixed em.


MoyTW (approved):
lol that's a big diff in the roomdata

PR #813, opened 2023-10-15
Fix move desyncs for multiple clients

MoyTW:
Still keeps the HTTP request for moving rooms, but moves contents into PubSub so it reaches all clients.

Doesn't fix weird text desync issues but that's MUCH more minor.

beforemove
aftermove
withmap


lazerwalker (approved):

lazerwalker:
LGTM! Feel free to merge

PR #815, opened 2023-10-17
content for the content gods

nextlevelbanana:
- save room data from prod
- add confetti-js library to use in orbOfZot room
- add items, room badges, 1 more generator for the secret quest
- make allowedItems list case insensitive
- more eel modifiers
- many more items in the Seers catalog


MoyTW (approved):
lol there really is a js package for whatever random little thing isn't there

PR #817, opened 2023-10-19
Fix login nonresponse

MoyTW:
This is an insanely bonkers fix, I have no real theory on how it works other than mumble mumble lazy evaluation, and I'm not sure it'll work on prod because with no theory on how it works how can I say with confidence? but here we are.

Behavior with main deployed:
brokenlogin
brokenbroken

With fix deployed:
fixedlogin
fixedfixed

### DIFF
--- a/server/src/endpoints/connect.ts
+++ b/server/src/endpoints/connect.ts
@@ -61,13 +61,18 @@ const connect: AuthenticatedEndpointFunction = async (user: User, inputs: any, l
   }
 
   result.groupManagementTasks = [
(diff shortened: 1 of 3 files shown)


lazerwalker:
I also have no idea why this would work?! Maybe `setUpRoomsForUser` somehow mutates the user ID and room ID on the user object such that shifting how/when the async call blocks in the promise makes a difference??

Either way, LGTM, I agree this is unlikely to cause new failures and 🎉🎉🎉 if it fixes it.

lazerwalker (approved):

PR #845, opened 2024-09-03
Uploading roomData from previous saved state where I had fully tweake…

BaldSavant:
…d rooms to setup all required links. Also updating the map to the 2024 map with clickable areas defined (hopefully correctly, but will need testing).


MoyTW (approved):
merged main in favoring yours, lgtm

PR #868, opened 2024-10-16
2024 secret badges

nextlevelbanana:
Added custom badges for the secret rooms
Fixed the lint errors
Fixed a minor bug where confetti rooms wouldn't let you click on the links because the confetti was in the way


MoyTW (approved):
lgtm

PR #870, opened 2024-10-17
Move all to entryway button

MoyTW:
Moves everybody to the entryway. When I tested it it did not cause the invisible chat error, I don't know if it clears invisible chat errors if they're already extant. I think it _should_ but who knows?

In the case that it doesn't the worst that it does is traps the people in the fairgrounds with an invisible chat, but they have to move out sometime, which will in itself clear it, so it's not...terrible if that's the case.


lazerwalker:
LGTM!

What value is there in returning the number of seconds the operation takes? Not saying it needs to be removed, just curious.

PR #872, opened 2024-10-18
Add Sidebar Obelisk

lazerwalker:
This re-adds the 'obelisk' button to the sidebar, with a much more complex implementation:

It now opens `ObeliskView`, a bespoke view that's mostly copy/pasted from `NoteWallView`. All of the network actions in there now call obelisk-specific endpoints. 

When you open ObeliskView, it fetches the current obelisk notes and returns them to you, as well as subscribing you to a pubsub group that gets note wall changes for the obelisk. On close, it unsubs you.

The intent of this feature is that the actual obelisk in the obelisk room and the sidebar obelisk are kept in sync.
1. In Redis, there is only the note wall associated with the `obelisk` roomID, the sidebar doesn't exist
2. When you make a sidebar obelisk note (i.e. the obelisk endpoints), we also push out pubsub changes to people in the obelisk room
3. When you make a normal obelisk note (i.e. in the obelisk room), we also push out pubsub changes to people in the sidebar obelisk group (i.e. the note wall endpoints are special cased to send this out if your current roomId is obelisk)

**This has been tested locally in my local-server branch, but not on a real instance**. Will do after merging.


server/src/endpoints/obelisk/addObeliskNote.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!message || !id) {
+    return {
+      httpResponse: {
+        status: 500,
+        body: 'Include a post-it message and an ID!'
+      }
+    }
+  }
+
+  await DB.addRoomNote('obelisk', { id, message, authorId: user.id })
```
lazerwalker, line 17: I would be happier if we had a constant somewhere to represent the `obelisk` roomId everywhere I hardcode it, but sigh, it's two days before the conf.

lazerwalker:
Note for future-me: the version of this that has the extra plumbing needed to work with my local server (i.e. routes entries and a small bugfix to the websocket manager this revealed) lives on my laptop's `obelisk-unrebased` branch.

server/src/endpoints/roomNote/likeRoomNote.ts
```diff
@@ -0,0 +1,48 @@
+import { AuthenticatedEndpointFunction, LogFn } from '../../endpoint'
+import { User } from '../../user'
+import DB from '../../redis'
+
+const likeRoomNote: AuthenticatedEndpointFunction = async (user: User, inputs: any, log: LogFn) => {
+  const noteId = inputs.noteId
+  if (!noteId) {
+    return {
+      httpResponse: {
+        status: 500,
```
MoyTW, line 10: probably should be 400 if malformed request?

server/src/endpoints/obelisk/startObservingObelisk.ts
```diff
@@ -0,0 +1,24 @@
+import DB from '../../redis'
+import { User } from '../../user'
+import { AuthenticatedEndpointFunction, LogFn } from '../../endpoint'
+
+const startObservingObelisk: AuthenticatedEndpointFunction = async (user: User, inputs: any, log: LogFn) => {
+  // The DB just knows there's a single obelisk, and doesn't distinguish between sidebar and the room
+  const notes = await DB.getRoomNotes('obelisk')
+
+  return {
+    groupManagementTasks: [
```
MoyTW, line 10: OK SO we have a bug in authenticatedAzureWrap, see 

https://github.com/Roguelike-Celebration/azure-mud/blob/fe8047ad8093a79d5e2e2a66be0294dd1f5ac79c/server/src/azureWrap.ts#L48-L49

https://github.com/Roguelike-Celebration/azure-mud/blob/fe8047ad8093a79d5e2e2a66be0294dd1f5ac79c/server/src/azureWrap.ts#L75-L76

The bug is that if you have no messages groupManagementTasks 500s because it's trying to push onto an array that's only created if we have messages, previously this must have been a non-issue because we only used this when we were notifying users anyways but it's an issue now, so either the bug needs to be fixed or `messages: []` needs to get added.

server/src/endpoints/roomNote/addRoomNote.ts
```diff
@@ -28,6 +28,18 @@ const addRoomNote: AuthenticatedEndpointFunction = async (user: User, inputs: an
       status: 200
     }
   }
+
+  // The sidebar obelisk and obelisk room share a noteWall, but have different pubsub groups to notify
+  // Adding an obelisk note from the sidebar is a different function, this is just the obelisk room case
+  if (user.roomId === 'obelisk') {
+    result.messages.push({
+      groupId: 'sidebar-obelisk',
```
MoyTW, line 36: If we ever have a room named sidebar-obelisk this will be a problem innit

MoyTW (changes requested):
You get 500s 'cuz of group management issues probably fine otherwise maybe? NEEDS user testing tho

server/src/endpoints/obelisk/startObservingObelisk.ts
MoyTW: fixed in latest commit

MoyTW (approved):

PR #897, opened 2025-04-13
License: fix author name, add years after 2020, and note in README

apjanke:
I don't think Microsoft wrote this code. :) I just guessed at "Roguelike Celebration Contributors" as a suitable collective author name.

Added years where more code was written to the copyright notice.

Noted the licensing type in the README. That text in `LICENSE` looks like the MIT License.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2020 Microsoft
+Copyright (c) 2020, 2021, 2022, 2023, 2024, 2025 Roguelike Celebration Contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


MoyTW (approved):
Oh, nice, thanks for checking that out.

Apparently, and I didn't know this, GitHub will autodetect common licenses?
Screenshot from 2025-04-16 16-36-05

apjanke:
> Apparently, and I didn't know this, GitHub will autodetect common licenses?

Yeah! That's how I figured it was MIT License in the first place: saw the little badge on the main GitHub repo page here, and then compared its text to https://opensource.org/license/MIT to confirm it. I don't know FLOSS licenses well enough to recognize them on sight, unless they like say "This is the GNU General Public License v3" in the first paragraph.

image

PR #908, opened 2025-05-13
Fix Web PubSub by ensuring PubSub user exists before AFA connect/ call

apjanke:
Addresses the https://github.com/Roguelike-Celebration/azure-mud/issues/899 breakage, using @lazerwalker's idea of pulling `conectPubSub()` up in `connect()`.

* Restore use of Web PubSub.
* Have networking.connect() call connectPubSub() before AFA connect/, to ensure the PubSub user exists.

If the PubSub user doesn't exist when connect/ is called, it may get error 500s from its azureWrap trying to do group-management things on the nonexistent user. Looks like it is the PubSub client connection creation that causes PubSub users to be auto-created. So do the connectPubSub() before calling any azureWrap-ed AFA functions, including connect/.

With this patch, on my test instance, the 500 errors are gone, I can see room content, and room chat between multiple users seems to be working. Whisper chat seems to work. Screenshot here shows fennel in Firefox and apjanke-1 in Chrome. 

image

### DIFF
--- a/server/src/azureWrap.ts
+++ b/server/src/azureWrap.ts
@@ -45,9 +45,6 @@ function outputToAzure (context: Context, req: HttpRequest, result: Result) {
   context.log('outputting to azure')
   context.log('context: ', JSON.stringify(context))
   context.log('result: ', JSON.stringify(result))
-
-  context.bindings.actions = []
-  /*
   if (result.messages) {
     context.bindings.actions = result.messages.map((m) => {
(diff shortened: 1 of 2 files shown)


MoyTW (approved):
Hey, you do the screenshot thing on PRs, huh. Nifty.

apjanke:
> Hey, you do the screenshot thing on PRs, huh. Nifty.

Wish everybody did. Let's try and make this A Thing. :)

PR #915, opened 2025-08-31
Update map size in MapView to fix click target locations

apjanke:
Fixes https://github.com/Roguelike-Celebration/azure-mud/issues/914 by updating the map size constants in click target area to match the new map size for 2025.

## References

This is followed up by https://github.com/Roguelike-Celebration/azure-mud/pull/916, which fixes this in a more robust manner by dynamically computing the map size instead of using constants. But that one has larger code changes, and I'm pretty sure this here PR just works.


MoyTW (approved):
Good eye for detail, I definitely didn't notice that.

PR #916, opened 2025-08-31
Dynamically compute ASCII map size in MapView click target logic

apjanke:
Avoids the need to update the map height & width constant when changing the mapText definition, by just computing them from `mapText` instead of using separate constants.

Fixes https://github.com/Roguelike-Celebration/azure-mud/issues/914 in a more robust way than https://github.com/Roguelike-Celebration/azure-mud/pull/915 which just updates the constant values.


MoyTW (approved):
Sensible.

PR #926, opened 2025-09-06
Fix generator links and handle non-loaded-room complex link targets

apjanke:
Fixes https://github.com/Roguelike-Celebration/azure-mud/issues/925.

Fixes the broken generator links for Tam and Madam Chrysalia by adding `linkActions` items for them. Looks like that is required to make generators actually usable as link targets. For the linkActions names, I used just `dromadTam` and `madamChrysalia` to match the links currently used in `roomData.json`, instead of the `talkToSoAndSo` style link action names that existing similar items were using.

Also changes the link handling logic in RoomView to handle additional cases like a room not being loaded to roomData yet (which is now normal because room data is JIT) or a bad link target name/ID, by having the "complex links" case default to rendering a room link, instead of degenerating to just "undefined".

Removes the "Dev warning: tried to link to room X, which doesn't exist" warnings around these cases, since the no-room-data case is now normal and common. Those warnings no longer indicate anything wrong, so I think are just a distraction. See https://github.com/Roguelike-Celebration/azure-mud/issues/919.


src/linkActions.ts
```diff
@@ -66,6 +66,12 @@ export const linkActions = {
   talkToFlower: () => {
     displayMessageFromList('flower')
   },
+  dromadTam: () => {
```
MoyTW, line 69: nit: Ideally we should remove older links too, because we're just kinda accumulating them.

src/components/RoomView.tsx
```diff
@@ -224,38 +224,32 @@ function parseDescription (
...
-      const userCountString = userCount > 0 ? ` (${userCount})` : ''
-      return `<a class='room-link' href='#' data-room='${roomId}'>${text}${userCountString}</a>`
     } else if (linkActions[roomId]) {
       return `<a class='room-link' href='#' data-action='${roomId}'>${text}</a>`
     } else {
-      // TODO: This warning is now expected, with room data being JIT
-      console.log(
-        `Dev warning: tried to link to room ${roomId}, which doesn't exist`
-      )
+      // With room data being JIT, it's now expected that roomData[roomId] might be empty,
```
MoyTW, line 242: Technically you can have rooms which don't exist in the roomData json but which are valid on the server, but I think in practice we'd want them to be congruent. The behavior of deferring to the server is correct, but I think that for testing purposes the dev warning is still helpful.

MoyTW (approved):
Minor nitpicks, will merge if you haven't enough time to address.

PR #930, opened 2025-09-06
Add 2025 preview schedule

MoyTW:


src/components/ScheduleView.tsx
```diff
@@ -17,7 +17,7 @@ export interface ScheduleEntry {
   day: number
 }
 
-const PREVIEW_END = new Date('2025-09-10T15:30:00-07:00')
+const PREVIEW_END = new Date('2025-09-10T17:30:00-07:00')
```
apjanke, line 20: image

I think the Preview date is 9/7, not 9/10 (which was just something I threw in as a guess earlier, I think).

PR #934, opened 2025-09-07
Minor JSON syntax fix for roomData.json

apjanke:
Remove extraneous trailing comma in roomData.json which may be technically invalid and cause some stricter JSON parsers to complain or fail.

See: https://github.com/Roguelike-Celebration/azure-mud/issues/932#issuecomment-3263503559


apjanke:
Closing since the new 2026 room data went in and this is stale now.

PR #947, opened 2025-10-21
Update Steam sale link for 2025

apjanke:
Update Steam sale [link for 2025](https://discord.com/channels/757389683188695150/760888641681358899/1429972403999735911) in room data. (The existing link says 2025, but is in the wrong format, so just goes to the Steam main page.)


MoyTW (approved):
yep link seems to work

PR #953, opened 2025-10-24
Fix extra periods in tarotPull gen, and add Fried Snacks gen

apjanke:
Fix errant punctuation in the tarotPull action, and including a very important generator I forgot: Fried Snacks from Under The Stage!

This PR is @BaldSavant's content, just with git fiddling from @apjanke.


apjanke:
Merged this, and did a prod deploy to pick it up: https://github.com/Roguelike-Celebration/azure-mud/actions/runs/18765517994

PR #955, opened 2025-10-24
Adding mysterious chess pieces to various rooms.

BaldSavant:

PR #956, opened 2025-10-24
Add mysterious chess pieces, fix M. Chrysalia "((expression))"

apjanke:
This is @apjanke's git tweaking of @BaldSavant's original #955 PR. It rearranges the commits to retroactively fix capitalization of "Pawn" in the first commit...

image

instead of doing it as a later separate bugfix commit like this...

image

Content of the PR is otherwise unchanged.

Supersedes and closes #955, which is the original variant of this.


apjanke:
Merged. Deployed to prod in https://github.com/Roguelike-Celebration/azure-mud/actions/runs/18767303318.

PR #963, opened 2025-10-24
Update currentEventBadge for 2025 event

apjanke:
Update the current event badge to "10" for 2025. It was left on last year's badge for 2024, and still was handing those out. Probably didn't catch this during the Preview because the current-event badge handout is time-gated, and didn't get activated until October; the Preview was back in September.

Addresses https://github.com/Roguelike-Celebration/azure-mud/issues/948, specifically https://github.com/Roguelike-Celebration/azure-mud/issues/948#issuecomment-3445095335.

### DIFF
--- a/server/src/badges.ts
+++ b/server/src/badges.ts
@@ -1,6 +1,9 @@
 import { keyBy } from 'lodash'
 import { BadgeCategories } from './types'
 
+// See server/src/moveToRoom.ts for the badge-awarding logic, including the setting for
+// which is the current-event badge.
+
 export interface Badge {
   /** We don't do any checks around string length, because Unicode Is Weird.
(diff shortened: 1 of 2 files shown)


MoyTW (approved):
I approve but the linter does not. Not worth blocking on it though.

apjanke:
To test, after this was deployed, I created a new rlc+yoda@apjanke.net user in the prod MUD, signed in, and moved around a bit. It got the "10" 2025 badge, and not the old "9" 2024 badge. Looks fixed.

image

The users which have been in the space so far this month - which doesn't include the Preview - will still have gotten the incorrect "9" 2024 event badge. Hopefully that's not too many.

PR #969, opened 2025-10-25
Fix Steam Sale link in nav bar for 2025 event

apjanke:
Link was updated in the gameTables room data, but not here.


apjanke:
Checks failed just due to a typecheck. I think that's unrelated; all checks are failing now like this bc our version of something doesn't support referrerpolicy, which was recently introduced to the embedded YT player.

image

PR #972, opened 2025-10-25
Add Pop Out Stream and Pop Out Captions links back to theater room description

### LINKED ISSUE #971: Missing "Pop Out Stream" links in theater
From Travis:

> There used to be a pop out stream in the theater which would open a window with a Twitch stream and the captions, it appears to be gone now.

## Diagnosis

In older versions of `roomData.json`, like from the 2024 event or [here](https://github.com/Roguelike-Celebration/azure-mud/blob/83179e1bab8bda41ce42b44aa16a4c1c281adeb5/server/src/rooms/data/roomData.json), there's code in the `theater` room description for it.

From d1fa2d7b, I see, in the end of the theater description:

```
(Check the \"Happening Now\" button on the left for speaker room assignments!)<br/><br/><a href=\"stream.html\" onClick=\"window.open('stream.html#' + window.getComputedStyle(document.body).getPropertyValue('background-color'), 'stream', 'width=560,height=460'); return false\">Pop Out Stream</a>. <a href=\"https://www.streamtext.net/player?event=RoguelikeCelebration \" target=\"_blank\">Pop Out Live Captions</a>.<br/>"
```

Image

Looks like you could just grab that and add it to the end of the theater description?

It uses `stream.html`. Doesn't look like we'll need to update that, because it doesn't have YT stream IDs in it, just a link to our Twitch channel and streamtext, which I think are still the same this year.

Image

## Possible fix

Here's the code

```
(Check the \"Happening Now\" button on the left for speaker room assignments!)<br/><br/><a href=\"stream.html\" onClick=\"window.open('stream.html#' + window.getComputedStyle(document.body).getPropertyValue('background-color'), 'stream', 'width=560,height=460'); return false\">Pop Out Stream</a>. <a href=\"https://www.streamtext.net/player?event=RoguelikeCelebration \" target=\"_blank\">Pop Out Live Captions</a>.<br/>
```

apjanke:
Fixes #971.

This PR gets the Pop Out Stream links etc. in to `roomData.json`, so if we do live editing to fix it in the theater now, it doesn't get lost if we revert to room data from disk.

### DIFF
--- a/server/src/rooms/data/roomData.json
+++ b/server/src/rooms/data/roomData.json
@@ -62,7 +62,7 @@
     "displayName": "The Big Top",
     "shortName": "big top",
     "id": "theater",
-    "description": "You enter the central tent to find a massive stage suffused with the smell of unseen fried snacks. Performers circle the edges, and even more strangely the attention of everyone in the room appears to be focused on people who are just standing in the center. Talking. You decide to sit and listen for a while.<br/><br/>You can return to the [[Pavilion]]. Or if you'd like to speak to one of our speakers after their talk, you can head to breakout rooms: [[Dressing Rooms]], [[Props Closet]], [[Break Room]], [[Under The Stage]]. (Check the 'Happening Now' button on the left for speaker room assignments!)",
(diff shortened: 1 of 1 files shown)


apjanke:
The check failure for this PR is just that broken typecheck.

image

apjanke:
Revised the PR just now to not put `<br>`s before those Pop Out links, to avoid pushing the text down.

Live edited the theater text in the room editor just now to use the same modification. Works for me, in Chrome on Mac.

image

MoyTW (approved):
lgtm

apjanke:
Ha, I _thought_ that preemptively merging this might come in handy. We _did_ do a Reset from Disk! (Accidentally.)

https://discord.com/channels/757389683188695150/889709795102326814/1432140271730163733

Suck it, skeptics.

PR #985, opened 2026-09-12
Preview event updates

apepers:
Fixed the URL for the theatre to point at the preview event, and took out reference to Twitter


apjanke (approved):
LGTM.

PR #986, opened 2026-09-13
Preview event fixes

ollyau:
Make gift shop accessible from map and amend the welcome modal to remove the obsolete video chat reference


apjanke (approved):
Lint is fixed. Looks good now. Merging.

